PR #1, opened 2020-02-28
kernel_settings role/module initial commit

richm:
This is the initial commit of the kernel_settings role/module


richm:
finally got travis tests to pass . . . please review, and if you want me to squash the above commits, please let me know

tyll:
It seems like this PR should also delete the `deleteme` file. If you need a file for the initial commit, another approach would be to use an empty `.gitignore` file which can be kept or updated in the future.

.travis.yml
```diff
@@ -2,9 +2,12 @@
 ---
 language: python
 env:
+# tuned cannot run in a container e.g. pytest unit tests of the module will fail
+# so use RUNNING_IN_CONTAINER=true so some tests can be skipped
   global:
     - LSR_ANSIBLES='ansible==2.6.* ansible==2.7.* ansible==2.8.*'
     - LSR_MSCENARIOS='default'
+    - RUNNING_IN_CONTAINER=true
```
pcahyna: Travis tests run in a container? (Molecule uses containers, but where are containers being used in Travis outside Molecule?)

pcahyna:
So far my main comment is that there are many Travis/Tox changes which should ideally be part of the template, so that other roles can benefit from them.

.travis.yml
richm: Maybe I don't understand how travis works then - because I was able to reproduce the error I see in travis when it runs pytest, by running pytest locally in a ubuntu xenial container.

richm:
> So far my main comment is that there are many Travis/Tox changes which should ideally be part of the template, so that other roles can benefit from them.

Definitely.  Once they are approved here, I will add them to the template, then to the other roles (and I still have a pending task to convert all of the roles to use tox . . .)

vars/default.yml
```diff
@@ -0,0 +1,6 @@
+---
+# Put internal variables here with values for unrecognized distributions.
+
+# Example:
```
pcahyna: I think you should delete the `# Example` line. It is not an example anymore, but an actual working code.

.travis.yml
pcahyna: @tyll @i386x can you help please? I am not that familiar with Travis myself.

args-operations.json
```diff
@@ -0,0 +1,33 @@
+{
```
pcahyna: what is this file?

args.json
```diff
@@ -0,0 +1,16 @@
+{
```
pcahyna: And this? From the documentation above, it seems to be related to tests, but shouldn't it belong to the tests directory then?

README.md
```diff
@@ -0,0 +1,260 @@
...
+
+Some parts related to `tuned` are `GPLv2+`.  These are noted in the headers
+of the files.  Everything else is `MIT`, except where noted.  See the file
+`LICENSE` for more information.
+
+## Author Information
+
+Rich Megginson (richm on github, rmeggins at my company)
+
+## Testing
```
pcahyna: While it is certainly useful to have testing docs, I don't think they should be part of the user docs (README.md) as the audience is different. @t-woerner any opinion?

defaults/main.yml
```diff
@@ -0,0 +1,31 @@
...
+# and to remove the entire sysfs section:
+# kernel_settings_operations:
+#   sysctl:
+#   - {"del": "kernel.threads-max"}
+#   - {"del": "fs.file-max"}
+#   sysfs: del
+kernel_settings_operations: {}
+# If force is true, completely wipe out whatever the current settings
+# are and replace them with kernel_settings_parameters
+kernel_settings_force: false
```
pcahyna: I know this has a huge bikeshed potential, but I don't think "force" is an appropriate name. To me, "force" sounds like "override some checks", not "wipe everything that's there".

install_ansible_for_testing.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+# this is used to install ansible into a venv or tox env
+# for testing purposes
```
pcahyna: I think it should be moved into some subdirectory (like .travis/ or tests/) if it is only testing-related. Ditto for the next file.

defaults/main.yml
pcahyna: FYI the selinux role uses `_purge` variables for that - check the docs. But we were not sure whether it is actually a good idea....

args-operations.json
richm: removed

args.json
richm: removed

README.md
richm: moved to README-testing.md

defaults/main.yml
richm: using `purge`

richm:
@pcahyna thanks - I think I have addressed your review feedback

richm:
[citest]

richm:
[citest bad]

richm:
[citest]

richm:
[citest pending]

richm:
[citest pending]

richm:
[citest]

richm:
[citest]

richm:
[citest]

richm:
[citest pending]

.travis.yml
tyll: @richm do you see the error also when you run in a xenial VM? The default for Travis nowadays is to run a VM not a container. Therefore only molecule should be using containers at the moment.

richm:
[citest]

richm:
> > So far my main comment is that there are many Travis/Tox changes which should ideally be part of the template, so that other roles can benefit from them.
> 
> Definitely. Once they are approved here, I will add them to the template, then to the other roles (and I still have a pending task to convert all of the roles to use tox . . .)

@pcahyna I have reverted almost all of these, and moved all of the customizations to `.travis/config.sh` which I think is the intention of that file.

pcahyna:
> > > So far my main comment is that there are many Travis/Tox changes which should ideally be part of the template, so that other roles can benefit from them.
> > 
> > 
> > Definitely. Once they are approved here, I will add them to the template, then to the other roles (and I still have a pending task to convert all of the roles to use tox . . .)
> 
> @pcahyna I have reverted almost all of these, and moved all of the customizations to `.travis/config.sh` which I think is the intention of that file.

@i386x any opinion on this please?

richm:
[citest]

richm:
[citest pending]

richm:
[citest]

tox.ini
```diff
@@ -0,0 +1,347 @@
...
+passenv = RUN_PYLINT_*
+setenv =
+    {[base]setenv}
+deps =
+    {[testenv]deps}
+    colorama
+    pylint>=1.8.4
+    ansible
+    -rpylint_extra_requirements.txt
+    -rtuned211_requirements.txt
```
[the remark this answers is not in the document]
richm: I can't, because I want to use tuned-2.11.0 for python 2.x (roughly corresponds to el7) and tuned-2.13.0 for python 3.x (roughly corresponds to fedora and el8)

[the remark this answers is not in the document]
richm: and if I have separate files for that anyway (I need tuned for pyunit as well as pylint) I might as well keep them in separate files rather than duplicate stuff in pylint_extra_requirements.txt

.travis.yml
```diff
@@ -0,0 +1,29 @@
+# SPDX-License-Identifier: MIT
+---
+language: python
+env:
+  global:
+    - LSR_ANSIBLES='ansible==2.6.* ansible==2.7.* ansible==2.8.*'
```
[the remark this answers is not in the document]
richm: I will revert molecule to `molecule<3`

molecule/default/molecule.yml
```diff
@@ -0,0 +1,34 @@
+---
+dependency:
+  name: galaxy
+driver:
+  name: docker
+lint: |
+  set -e
```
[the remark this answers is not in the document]
richm: I'm reverting back to `molecule<3`

richm:
@i386x Thanks for the review - I think I have addressed your comments - please check

richm:
> `     - RUNNING_IN_CONTAINER=true `

The issue wasn't running in a container, the issue was running on ubuntu, which I solved with some judicious monkeypatching

pcahyna:
> > `    - RUNNING_IN_CONTAINER=true`
> 
> The issue wasn't running in a container, the issue was running on ubuntu, which I solved with some judicious monkeypatching

would running even the unit test (assuming that it were the unit test that were the problem, not Molecule tests...) in a container (CentOS or Fedora) help? If the problem is with the Ubuntu kernel, it won't help, but if it is with the userland (like tuned) it should.

richm:
> > > `    - RUNNING_IN_CONTAINER=true`
> > 
> > 
> > The issue wasn't running in a container, the issue was running on ubuntu, which I solved with some judicious monkeypatching
> 
> would running even the unit tests (assuming that it were the unit tests that were the problem, not Molecule tests...) in a container (CentOS or Fedora) help? If the problem is with the Ubuntu kernel, it won't help, but if it is with the userland (like tuned) it should.

For unit testing, it was as simple as this to make it work on ubuntu: https://github.com/linux-system-roles/kernel_settings/pull/1/files#diff-ca666a588d70612a44b9f4ef999c9525R798
```python
    orig_set_receive_buffer_size = pyudev.monitor.Monitor.set_receive_buffer_size
    pyudev.monitor.Monitor.set_receive_buffer_size = lambda self, size: None
```

README.md
```diff
@@ -0,0 +1,223 @@
...
+  sysctl:
+    kernel.threads-max: 29968
+    vm.max_map_count: 65530
+    fs.epoll.max_user_watches: 785592
+    fs.file-max: 379724
+```
+This allows multiple higher level roles or playbooks to use this role to
+provide the kernel settings specific to that component.  For example, if you
+are installing a web server and a database server on the same machine, and
+they both require setting kernel parameters, the `kernel_settings_parameters`
```
olysonek: You're talking about the role itself here. I think `kernel_settings` was correct, wasn't it?

richm: yes

library/kernel_settings.py
```diff
@@ -0,0 +1,990 @@
...
+            - Remove the current kernel_settings, whatever they are, and force
+              the given settings to be the current and only settings
+
+author:
+    - Rich Megginson (rmeggins@redhat.com)
+"""
+
+EXAMPLES = """
+# add the sysctl vm.max_mmap_regions, and disable spectre/meltdown security
+kernel_settings:
```
olysonek: `kernel_settings_parameters`?

library/kernel_settings.py
```diff
@@ -0,0 +1,990 @@
...
+        /sys/kernel/debug/x86/pti_enabled: 0
+        /sys/kernel/debug/x86/retp_enabled: 0
+        /sys/kernel/debug/x86/ibrs_enabled: 0
+
+# replace the existing sysctl section with the specified section
+# delete the /sys/kernel/debug/x86/retp_enabled setting
+# completely remove the vm section
+# add the bootloader cmdline arguments spectre_v2=off nopti
+# remove the bootloader cmdline arguments quiet splash
+kernel_settings:
```
olysonek: `kernel_settings_parameters`?

library/kernel_settings.py
```diff
@@ -0,0 +1,990 @@
...
+    sysctl:
+        vm.max_mmap_regions: 262144
+    bootloader:
+        cmdline: spectre_v2=off nopti
+    operations:
```
olysonek: This (and also the `DOCUMENTATION` variable above) suggests that the `operations` are on the same level as `sysctl` etc. Whereas the README suggests that the `operations` are on the same level as `kernel_settings_parameters`. Is there an inconsistency in the docs, or am I misunderstanding something? (I'm not really familiar with ansible, so that's a real possibility.)

richm:
[citest]

library/kernel_settings.py
richm: No, this is for the module, not the role.  The name of the module is `kernel_settings`.

library/kernel_settings.py
richm: module name

library/kernel_settings.py
richm: This documentation is for the module, not the role.  The module is not intended to be used outside of the role.  I guess you could say the role is the "interface" and the module is the default "implementation" of the role interface.

richm:
[citest]

richm:
[citest]

richm:
[citest]

richm:
[citest]

richm:
[citest]

richm:
[citest pending]

richm:
[citest]

richm:
[citest pending]

olysonek:
Hi, a full review is still coming. In the meantime, I thought we could discuss the fact that the code may break when Tuned is refactored, or just the internal API changes for some reason. Do you have a plan how to handle that?

I'd suggest writing tests for tuned for all the behaviours that you depend on. In the spirit of Google's codebase, "if you like it, you should put a test on it" ;). That way, we'll know soon enough that we're breaking something that the kernel_settings role depends on.

Here's an example of such a test. (It doesn't work yet in our CI, though.)
https://github.com/redhat-performance/tuned/pull/258/files

richm:
> Hi, a full review is still coming. In the meantime, I thought we could discuss the fact that the code may break when Tuned is refactored, or just the internal API changes for some reason. Do you have a plan how to handle that?

I'd like to get to the point where I can just use the `tuned-adm` or `tuned` command directly without having to "link" to the code, sort of like how the OpenShift machine tuning operator is using tuned.  In the meantime, I'll just have to keep up with tuned development in case there is some release that changes an internal api.  I don't expect any changes in RHEL7 in that respect, so mostly focused on rhel8 and fedora.  I may have to create a CI job that runs periodically against tuned master branch or release branches (if any), or perhaps an internal only CI job that pulls the source code from rh dist git rhel-7.latest and rhel-8.latest branches (because the internal branches may have backported code from upstream master).

So far, I have been able to create the kernel_settings code to handle multiple versions of tuned - I had to write it to handle both tuned-2.10 and tuned-2.11 which did have some api changes I had to take into consideration.  For example: https://github.com/linux-system-roles/kernel_settings/pull/1/files#diff-ca666a588d70612a44b9f4ef999c9525R193 - the `load_all_plugins` method is not available in 2.10, so I just hardcoded the list of plugins if that method is not available.

> 
> I'd suggest writing tests for tuned for all the behaviours that you depend on. In the spirit of Google's codebase, "if you like it, you should put a test on it" ;). That way, we'll know soon enough that we're breaking something that the kernel_settings role depends on.

Ok.  I'll do that, and in the meantime, if you spot something I should have a test for, let me know.

> 
> Here's an example of such a test. (It doesn't work yet in our CI, though.)
> https://github.com/redhat-performance/tuned/pull/258/files

Ah - this looks like it is related to what I was talking about above with `load_all_plugins`

olysonek:
Ok, thanks. That sounds reasonable.

> > Here's an example of such a test. (It doesn't work yet in our CI, though.)
> > https://github.com/redhat-performance/tuned/pull/258/files
> 
> Ah - this looks like it is related to what I was talking about above with `load_all_plugins`

Yes, that's a test for one of the things the kernel_settings role depends on.

yarda:
For the bootloader you can use the following Tuned syntax:
For option addition:
[bootloader]
cmdline_unique_suffix1=+OPT1

For option removal:
[bootloader]
cmdline_unique_suffix1=-OPT1

For replacing all previously set cmdline by Tuned:
[bootloader]
cmdline=OPT

Due to the historical limitation of the used parser, the '_unique_suffix...' needs to be used - you can use any string, but it needs to be unique across all profiles loaded at once - we usually use _PROFILE_NAME there. If you use non unique suffix, standard merge operation of the option will be done first, i.e.:
cmdline1=+OPT1
cmdline1=+OPT2
Standard Tuned options merge will happen and just the 'OPT2' will be added

richm:
> I am not Ansible/Ansible roles devel, so I am taking the review from the Tuned side (please pardon my ignorance if I pointed out something which is OK/standard approach in the "Ansible way"). 

No problem.  I was primarily looking for comments about tuned and python usage.

> Comments follows:
> 
>     * Why are there different requirements for the tuned-2.11.0 (tuned211_requirements.txt) and tuned-2.13.0 (tuned213_requirements.txt)? I.e. different procfs versions from the git. IMHO Tuned shouldn't be sensitive on the procfs version (I cannot see anything procfs version specific in the tuned-2.11.0/2.13.0 diff). So I think there could be one requirement file and not three.

This is only for testing purposes (pytest and pylint), for testing in a tox venv with pip installed packages, and primarily for the upstream Travis CI which runs on an Ubuntu VM.

> 
>     * Is there any need for the deps to be installed through the pip? Isn't possible to use the standard RPM deps?

This is only for testing purposes (pytest and pylint), for testing in a tox venv with pip installed packages, and primarily for the upstream Travis CI which runs on an Ubuntu VM.

> 
>     * Tuned supports 'make install' and DESTDIR, so it's not needed to manually copy the files, standard 'make' could be used.

Not sure what you are you referring to here.

> 
>     * tuned-adm verify has the '-i' option which means ignore_missing (it is 'ignore_missing' bool option in the internal API, i.e. verify_profile(ignore_missing)). If enabled, the verification shouldn't treat most of the missing/non-supported findings as errors.

Ok.

> 
>     * In Tuned profiles the 'replace' option is supported, e.g.:
> 
> 
> ```
> [sysctl]
> replace=1
> ...
> ```
> 
> It means it will throw away all previously set plugin options (for the specific plugin) and will use just what defined here. So maybe you could utilize this feature and not re-implement it.

Ok.

richm:
> For the bootloader you can use the following Tuned syntax:

Thanks - I'll check it out

richm:
> Is there any need for the deps to be installed through the pip? Isn't possible to use the standard RPM deps?

We install the `tuned` package for the deps at runtime for real deployments, as opposed to unit test time:
https://github.com/linux-system-roles/kernel_settings/pull/1/files#diff-2444ad0870f91f17ca6c2a5e96b26823R12
```
- name: Ensure required packages are installed
  package:
    name: "{{ __kernel_settings_packages }}"
    state: present
```
https://github.com/linux-system-roles/kernel_settings/pull/1/files#diff-1190127f593c90fc6a128bf65eb4f39dR2
```
 __kernel_settings_packages: ["tuned"]
```

richm:
Thanks reviewers.  I have a new version for you to review.  I think I have addressed your review comments.  @yarda @olysonek I think your suggestions will be applicable if/when we move away from the python module approach and towards a more `tuned-adm` based implementation, where you would have multiple profiles (e.g. to replace a section in an included profile, to add/remove bootloader options in an included profile).
@pcahyna This implements the incremental changes proposal.  Please check the README.

richm:
Please review, or just comment `lgtm` if you have no further review suggestions.  I'd like to get this merged soon.

.travis/config.sh
```diff
@@ -69,6 +69,10 @@ function ks_lsr_install_tuned() {
 
 # install tuned only where needed
 KS_LSR_PYVER=$(lsr_get_python_version $(type -p python))
+if lsr_compare_versions $KS_LSR_PYVER -lt 2.7 ; then
+  lsr_info kernel_settings tests not supported on python $KS_LSR_PYVER - skipping
```
pcahyna, line 73: Could you please leave a comment why is it unsupported? Is it because the tests are unsupported on Python 2.6, or that the module code itself does not support Python 2.6 (I see that you exclude EL6 - the main reason why one wants to support Python 2.6 - here: https://github.com/linux-system-roles/kernel_settings/pull/1/files#diff-9b4ddf055111f4371f04fbb8b5366655R14 so it is probably the latter, but the wordings suggests the former.)

tests/tasks/assert_kernel_settings_values.yml
```diff
@@ -0,0 +1,61 @@
...
+    - __kernel_settings_register_parameters.bootloader is defined
+    - __kernel_settings_register_parameters.bootloader.cmdline is defined
+
+- name: verify bootloader settings command
+  set_fact:
+    __kernel_settings_success: false
+  when:
+    - __kernel_settings_register_verify_bl_cmdline is defined
+    - __kernel_settings_register_verify_bl_cmdline.rc is defined
+    - __kernel_settings_register_verify_bl_cmdline.rc != 0
```
pcahyna, line 52: there is a test for this, https://docs.ansible.com/ansible/latest/user_guide/playbooks_tests.html#task-results

pcahyna: but will it ever succeed? With a nonzero exist status, how will we reach this task?

README.md
```diff
@@ -0,0 +1,222 @@
...
+will result in the values `spectre_v2=off nopti quiet splash` being added to,
+or replacing, the existing `bootloader` `cmdline` values.  If you want to
+remove whatever was there, and replace it with your specified values, use
+`state: absent` in the value list:
+```yaml
+kernel_settings_parameters:
+  bootloader:
+    - name: cmdline
+      value:
+        - state: absent
```
pcahyna, line 151: I think we agreed to use `previous: replaced` for this, not `state: absent`. (`state: absent` would be strange, because by using it you are saying that the list should be absent, but after running the role, there is actually something in the list. `state` is the desired state, which is not `absent` in this case - you want something to be there.)

README.md
```diff
@@ -0,0 +1,222 @@
...
+This will remove the `vm.max_map_count` setting from the `sysctl` section.
+If you want to remove an entire section, specify `state: empty` as a `dict`
+instead of a `list`:
+```yaml
+  sysctl:
+    state: empty
+```
+This will have the effect of removing all of the `sysctl` settings.
+
+The `bootloader` `cmdline` value is a special case.  There may be situations
```
pcahyna, line 127: Why is it special? To me it looks similar to the above.

README.md
```diff
@@ -0,0 +1,222 @@
...
+where you want to treat it as a list of values, and add or remove specific
```
pcahyna, line 128: Is there a situation where you *don't* want to treat it as a list of values but as something else? It does not seem to be shown in this document.

README.md
```diff
@@ -0,0 +1,222 @@
...
+      value: 0
+    - name: /sys/kernel/debug/x86/retp_enabled
+      value: 0
+    - name: /sys/kernel/debug/x86/ibrs_enabled
+      value: 0
+  vm:
+    - name: transparent_hugepages
+      value: madvise
+  bootloader:
+    - name: cmdline
```
pcahyna, line 56: would it work to have it as a dict:
```yaml
bootloader:
  cmdline:
    - name: spectre_v2
      value: "off"
    - name: nopti
```
? It would simplify the syntax.

(For the settings above, there are name/value pairs, but those are different. For settings like sysctl, sysfs, the names are essentially arbitrary, constrained only by what the kernel supports. While `cmdline` under `bootloader` is, I suppose, a special entry. You will not have a list of multiple differently-named cmdlines. So it makes sense to make `cmdline` a dictionary key, not a value of `name`.)

README.md
pcahyna: (except that it is nested one level below bootloader, while in case of the above they are on the toplevel. But otherwise `cmdline` seems to behave in a pretty similar way as `sysctl` does.)

.travis/config.sh
richm: The main problem is that not all of the dependencies for running tuned using python 2.6 are available via pypi.  For example, you cannot install `PyGObject` from pypi - none of the versions available on pypi will install on python 2.6.

tests/tasks/assert_kernel_settings_values.yml
richm: > but will it ever succeed? With a nonzero exist status, how will we reach this task?

Not sure what you mean?

README.md
richm: The `bootloader` `cmdline` value is a list.  Every other value at the same level is a scalar.

README.md
richm: I will change the wording.  It is always a list.

README.md
richm: I think the disadvantage of making it inconsistent with the other settings, and the potential confusion caused, outweigh any advantage.

richm:
Thanks @pcahyna - I think I have addressed all of your review comments.

tests/tasks/assert_kernel_settings_values.yml
pcahyna: `__kernel_settings_register_verify_bl_cmdline` comes from a previous `command` task 
The previous task does not have any `ignore_errors` or `failed_when`.
Therefore, if `rc` is nonzero, it will fail and we will not reach this task. The test `__kernel_settings_register_verify_bl_cmdline.rc != 0` will never succeed. Am I missing something?

library/kernel_settings.py
```diff
@@ -0,0 +1,853 @@
...
+    plugins_repo = tuned.plugins.repository.Repository(
+        None, None, None, None, None, None, None, None
+    )
+    for plugin_class in plugins_repo.load_all_plugins():
+        try:
+            tmpplg = plugin_class(
+                None, storage_factory, None, None, None, None, None, None
+            )
+            names.append(tmpplg.name)
+        except NotSupportedPluginException:
```
olysonek, line 276: The question is, do you want to list unsupported plugins?

README-testing.md
```diff
@@ -0,0 +1,82 @@
...
+            {"name": "cmdline", "value": [
+		        {"name": "mitigations", "value": "on"},
+		        {"name": "another"}
+	            ]
+	        }
+        ],
+        "selinux": [
+            {"name": "avc_cache_threshold", "value": 512}
+        ],
+        "state": "present",
```
olysonek, line 69: Should the "state" key be here? Looking at the code, it doesn't seem to be recognized. And running the module with this args.json yields the following:
```
{
    "msg": "Invalid format for input parameters",
    "warnings": [
        "Error: section [state] value must be a dict or a list"
    ],
...
```

olysonek: Actually there is `state=dict(type="str", required=False),` in `module_args`. But it doesn't seem to work. Is this some leftover?

tests/tasks/assert_kernel_settings_values.yml
richm: Got it - fixed it - also converted the test to work with the bootloader value being a `list` instead of a `dict` (that is, there is no more `parameters.bootloader.cmdline`) - see https://github.com/linux-system-roles/kernel_settings/pull/1/commits/0ac194a33e6976d11a7baae252e848797f875a21

library/kernel_settings.py
richm: Yes, I think so, if possible.  For example - the user provides a single set of values for multiple platforms, and the kernel_settings/tuned module applies only those plugins that are supported on the specific platform, and ignores the other plugins.

If this isn't possible, then the user will have to set platform specific kernel settings - also doable, just not as elegant.

README-testing.md
richm: > Is this some leftover?

Yes - removed in latest commit - https://github.com/linux-system-roles/kernel_settings/pull/1/commits/712eef3156ed9ed38236a66a55b14d05e3d0bb8e

library/kernel_settings.py
```diff
@@ -0,0 +1,853 @@
...
+    return tuned_config
+
+
+def load_current_profile(tuned_config, tuned_profile, logger):
+    """load the current profile"""
+    tuned_app = None
+    errmsg = "Error loading tuned profile [{}]".format(TUNED_PROFILE)
+    try:
+        tuned_app = tuned.daemon.Application(tuned_profile, tuned_config)
+        if tuned_app is not None and tuned_app.daemon is not None:
```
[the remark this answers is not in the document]
richm: are you saying that on line 645 it is not possible for `tuned_app` to be `None`?  Or that it is not possible for `tuned_app.daemon` to be `None`?

library/kernel_settings.py
```diff
@@ -0,0 +1,853 @@
...
+    if "TEST_ROOT_DIR" not in os.environ:
+        atexit.register(test_cleanup)
+    return test_cleanup
+
+
+def get_tuned_config():
+    """get the tuned config and set our parameters in it"""
+    tuned_config = tuned.utils.global_config.GlobalConfig()
+    tuned_config.set("daemon", 0)
+    tuned_config.set("reapply_sysctl", 0)
```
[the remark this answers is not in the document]
richm: > Posting some minor comments...
> 
> One thing that just occured to me: instead of using the internal tuned API, couldn't you just do something along the lines of the following?
> 
> ```
> < update the kernel_settings profile according to the module parameters >
> 
> if ! grep kernel_settings /etc/tuned/active_profile; then
>     tuned-adm profile kernel_settings $(tuned-adm active)
> else
>     systemctl restart tuned
> fi
> ```
> 
> (Even the RHEL-7 Tuned seems to support applying multiple profiles.) Sorry if this has been raised before.

Not sure what you mean.  Do you mean to implement the kernel_settings module without the python wrapper around the tuned code e.g. as a series of ansible tasks that invoke modules such as `ini_file`, and shell commands?  If so, I think that is something we should do in the future, if possible.

Or do you mean specifically the bash code excerpt above about managing the active profile and restarting tuned, and that we could do that instead of https://github.com/linux-system-roles/kernel_settings/pull/1/files#diff-ca666a588d70612a44b9f4ef999c9525R495 ?  If so, I'm not sure what that would gain us.  Also note that in ansible, the restart of tuned is conditional.  If the module detects that something has changed, it will return a value of `"changed":true` in the result: https://github.com/linux-system-roles/kernel_settings/pull/1/files#diff-ca666a588d70612a44b9f4ef999c9525R833
The task will notify the handler if the module returns `"changed":true` https://github.com/linux-system-roles/kernel_settings/pull/1/files#diff-2444ad0870f91f17ca6c2a5e96b26823R38

library/kernel_settings.py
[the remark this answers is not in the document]
olysonek: I think neither is possible. But I may be wrong...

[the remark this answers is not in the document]
richm: Fixed - https://github.com/linux-system-roles/kernel_settings/pull/1/commits/5412313d92fbf08c9ee9b770a42ff4e1bbc53c1b - see also adding setting `dynamic_tuning` to `0` - https://github.com/linux-system-roles/kernel_settings/pull/1/commits/fb85df9d6ee690fc9a493153c445a4e298bd3a54

library/kernel_settings.py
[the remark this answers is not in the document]
olysonek: Actually, never mind. What I was really wondering is how is the "profile pre-load" feature (https://bugzilla.redhat.com/show_bug.cgi?id=1798183) is going to help you, given that there is already the possibility to apply multiple profiles. And I think the answer to that is that you want the `kernel_settings` profile to remain applied even if the user manually runs `tuned-adm profile <some other profile>`. Right?

library/kernel_settings.py
olysonek: Sounds OK to me. I just wanted to ask.

README-testing.md
olysonek: OK. Can you remove it from README-testing as well?

library/kernel_settings.py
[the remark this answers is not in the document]
richm: > And I think the answer to that is that you want the kernel_settings profile to remain applied even if the user manually runs tuned-adm profile <some other profile>. Right?

Right.  So for now, we are relying on the user not to remove the kernel_settings profile, or otherwise relying on the user to not change tuned config in such a way that kernel_settings are not applied.  Note that this is not a new problem, and not a problem with the kernel_settings role specifically (e.g. the network role - the user can go in and manually change network settings, outside of the role, that will cause the settings applied by the network role do not function correctly) - but we don't want it to be too easy for the user to alter the kernel_settings using `tuned-adm` or editing tuned profiles.

richm:
@olysonek @yarda @pcahyna Thanks for the review feedback.  See https://github.com/linux-system-roles/kernel_settings/pull/1/commits/338afd5695bdba22882a112fada3f3618432e3e8 which addresses some of the items.  @olysonek please confirm that this solves the problems with the use of `unit_manager.create` (which I removed) and with the storage files in /var/run/tuned and /var/lib/tuned.

I still have to change the interface to use `kernel_settings_sysctl` instead of `kernel_settings_parameters.sysctl` etc., and to improve the `README-testing.md` - @olysonek please let me know what problems you ran into trying to run/debug.

richm:
@pcahyna I think this latest commit https://github.com/linux-system-roles/kernel_settings/pull/1/commits/af7311dd24bb02e7d3bad7c5b7cbe29f3e28d566 implements what you were talking about.  One question is: How to delete the scalar settings such as `kernel_settings_systemd_cpu_affinity`?
- set to some special value e.g. `kernel_settings_systemd_cpu_affinity: absent`?  I'm not sure what would be an appropriate value
- set to the `{'state':'empty'}` value, and allow the value to be a `dict` or a scalar - seems inelegant
- do not set the value to `null` in the defaults/main.yml - have it be undefined - and use `null` as the deletion value - seems logical, but perhaps this goes against best practices and/or our conventions
```
kernel_settings_systemd_cpu_affinity is not defined == that setting is not touched
kernel_settings_systemd_cpu_affinity: null == that setting is removed from the existing config
kernel_settings_systemd_cpu_affinity: some_value == setting is set to that value
```
other ideas?

richm:
> other ideas?

Have a `NAME_state` parameter - if `NAME_state: absent` then remove `NAME` from the current settings.
```
kernel_settings_systemd_cpu_affinity_state: absent
kernel_settings_transparent_hugepages_state: absent
kernel_settings_transparent_hugepages_defrag_state: absent
```
I'm going to do a PoC of this approach.

richm:
@pcahyna https://github.com/linux-system-roles/kernel_settings/pull/1/commits/2938cc51fbe8ef65e026a294a3309f5ed4c7c023

tests/tasks/assert_kernel_settings_conf_files.yml
```diff
@@ -0,0 +1,92 @@
...
+  when: __kernel_settings_register_profile_conf_result is failed
+
+- name: get active_profile file
+  slurp:
+    src: /etc/tuned/active_profile
+  register: __kernel_settings_register_profile_conf_active_profile
+
+# that is, the value must be exactly 'kernel_settings'
+# or 'kernel_settings something'
+- name: check that active_profile starts with kernel_settings
```
olysonek, line 50: `s/starts/ends/` ?

richm: good catch

library/kernel_settings.py
```diff
@@ -0,0 +1,777 @@
...
+    tuned_app, errmsg = load_current_profile(tuned_config, TUNED_PROFILE, module)
+
+    if tuned_app is None:
+        module.fail_json(msg=errmsg, **result)
+    else:
+        current_profile = tuned_app.daemon.profile
+        debug_print_profile(current_profile, module)
+        errmsg = ""
+
+    # if the user is working with this module in only check mode we do not
```
olysonek, line 729: It's unclear to me whether the check mode does anything useful given the changes in 338afd5695b.

library/kernel_settings.py
```diff
@@ -0,0 +1,777 @@
...
+    profile_base_dir = tuned.consts.LOAD_DIRECTORIES[-1]
+    prof_fname = os.path.join(
+        profile_base_dir, TUNED_PROFILE, tuned.consts.PROFILE_FILE
+    )
+    with open(prof_fname, "wb") as prof_f:
+        cfg.write(prof_f)
+
+
+def update_current_profile_and_mode(daemon):
+    """ensure that the tuned current_profile applies the kernel_settings first"""
```
olysonek, line 426: `s/first/last/` ?

olysonek:
> please confirm that this solves the problems with the use of `unit_manager.create` (which I removed) and with the storage files in /var/run/tuned and /var/lib/tuned

Indeed it looks like the problem is not there (anymore). Thanks.

> please let me know what problems you ran into trying to run/debug

In light of what I learned during our meeting, I think these are non-issues:
1. When I tried to run the module using something like `python3 library/kernel_settings.py`, the module reported an error that the kernel_settings profile doesn't exist. But as you told me, creating the profile is handled outside the module, inside the playbook (or whatever it is called :)).
2. When I tried to run the module, I was surprised that it didn't actually do any tuning. I mistakenly thought that applying the tuning is done directly by the module. But as I learned, this is again handled in the playbook.

I think I'm done with the review now and have no further comments at this time :).

richm:
> Let me just say that the Tuned features `reapply_sysctl` and `dynamic_tuning` are a bit unfortunate for your use case. Imagine for example that a user has applied (outside of kernel_settings) a profile that doesn't use the `disk` nor the `sysctl` plugin. Then the user decides to set some sysctl settings and disk readahead using the `kernel_settings` role. The resulting `kernel_settings` profile could look something like this:
> 
> ```
> [sysctl]
> vm.swappiness=10
> 
> [disk]
> devices=sda
> readahead=4096
> ```
> 
> If `reapply_sysctl=1` is set, running the `kernel_settings` role will result in all sysctl settings from `/etc/sysctl*` being applied. If `dynamic_tuning=1` is set, running the role will result in the `sda` device being tuned dynamically depending on the current load. But the user didn't ask for any of this.
> 
> Note that `reapply_sysctl=1` is set by default in RHEL, Fedora and upstream. `dynamic_tuning=1` is set by default in Fedora and upstream.

I suppose that when the role installs `tuned` (as opposed to when the user has already installed and configured `tuned`), the role could set
```
reapply_sysctl=0
dynamic_tuning=0
```
That way, if the user doesn't do anything, they will get the behavior desired.

> 
> Sorry if this kind of feedback is coming late. My understanding of the role changed somewhat during our meeting last week :).

No worries.

richm:
Re: `reapply_sysctl` - I think it is ok if the default is `reapply_sysctl=1`.  If the user is using `kernel_settings` *and* using the `sysctl` module, or manually editing /etc/sysctl.conf or /etc/sysctl.d/ files, then it is the responsibility of the user to prevent conflicts.  The role does try to do some verification - after applying the settings, we restart tuned, then run `tuned-adm verify` - https://github.com/linux-system-roles/kernel_settings/pull/1/files#diff-3d2b7487e8f37272aeb6f25599f2734bR2

Presumably this will fail if the user has provided some conflicting settings:
- role sets sysctl settings
- role restarts tuned
- tuned applies profile settings, then reapply_sysctl with conflicting settings
- role runs tuned-adm verify - FAIL - the setting in the profile was overridden by sysctl file setting

I think we can only run into this situation if the user has explicitly made sysctl settings that conflict with the settings made with the `kernel_settings` module.  In that sense, is it no different than the user doing some other system or tuned configuration that counters what the role set up. 

Re: `dynamic_tuning` - for now, anyway, we are only supporting `sysctl`, `sysfs`, `vm`, `systemd`, `bootloader` (and maybe `modules`) - https://github.com/linux-system-roles/kernel_settings/pull/1/files#diff-ca666a588d70612a44b9f4ef999c9525R235 - I don't think any of these use dynamic tuning.

richm:
> > please confirm that this solves the problems with the use of `unit_manager.create` (which I removed) and with the storage files in /var/run/tuned and /var/lib/tuned
> 
> Indeed it looks like the problem is not there (anymore). Thanks.
> 
> > please let me know what problems you ran into trying to run/debug
> 
> In light of what I learned during our meeting, I think these are non-issues:
> 
>     1. When I tried to run the module using something like `python3 library/kernel_settings.py`, the module reported an error that the kernel_settings profile doesn't exist. But as you told me, creating the profile is handled outside the module, inside the playbook (or whatever it is called :)).

There is some setup required first, using `tox` - see https://github.com/linux-system-roles/kernel_settings/pull/1/files#diff-67385c2b6739d65912ec02759af26815R2
then, you have to set some env. vars. e.g.
```
$ TESTING=true [TEST_PROFILE=kernel_settings] python library/kernel_settings.py args.json
```
`TESTING=true` is what configures an `/etc/tuned` file/directory structure with the right files and the right profiles for testing (basically, copy `tests/tuned/` to a temp directory) - https://github.com/linux-system-roles/kernel_settings/tree/13d6040313d87a5df60f9a2235288ba013cf23b5/tests/tuned/etc/tuned - use `TEST_PROFILE=name` to use one of the other test profiles in `tests/tuned/etc/tuned/`

> 
>     2. When I tried to run the module, I was surprised that it didn't actually do any tuning. I mistakenly thought that applying the tuning is done directly by the module. But as I learned, this is again handled in the playbook.

Right.  The module only writes out the files, then restarts `tuned` - `tuned` then applies the settings when it starts.  In general, I would prefer that unit testing the module *not* change the kernel settings on the system - that should really only be done in the virtual machine based tests.

> 
> 
> I think I'm done with the review now and have no further comments at this time :).

library/kernel_settings.py
richm: fixed

library/kernel_settings.py
richm: good catch - fixed

olysonek:
> I suppose that when the role installs `tuned` (as opposed to when the user has already installed and configured `tuned`), the role could set
> 
> ```
> reapply_sysctl=0
> dynamic_tuning=0
> ```

If you can detect that case reliably, then great, you could do that. It might be a bit confusing that the role behaves differently in certain cases, though. Perhaps you could document this.

>  If the user is using `kernel_settings` _and_ using the `sysctl` module, or manually editing /etc/sysctl.conf or /etc/sysctl.d/ files, then it is the responsibility of the user to prevent conflicts.

The user doesn't have to manually edit /etc/sysctl.d/ files. The files could be installed by some system packages. On my Fedora, I have `/etc/sysctl.d/50-libreswan.conf`, owned by `libreswan-3.31-2.fc31.x86_64`.

>  The role does try to do some verification - after applying the settings, we restart tuned, then run `tuned-adm verify` - https://github.com/linux-system-roles/kernel_settings/pull/1/files#diff-3d2b7487e8f37272aeb6f25599f2734bR2
> 
> Presumably this will fail if the user has provided some conflicting settings

Yes, and that's a good thing. But I'm not really talking about conflicts here. I mean, conflicts can be a consequence of the current behaviour, sure. But what I'm saying is that the role will set more sysctl settings than the user explicitly asked it to do.

I'm not saying that this is necessarily a deal breaker, but it is something that you should be aware of and something that you might want to document.

> I think we can only run into this situation if the user has explicitly made sysctl settings that conflict with the settings made with the `kernel_settings` module.

Not really. See above.

> `dynamic_tuning` - for now, anyway, we are only supporting `sysctl`, `sysfs`, `vm`, `systemd`, `bootloader` (and maybe `modules`) - https://github.com/linux-system-roles/kernel_settings/pull/1/files#diff-ca666a588d70612a44b9f4ef999c9525R235 - I don't think any of these use dynamic tuning.

Good. Indeed, they don't use dynamic tuning. And I think I would limit the list of plugins to those that don't use dynamic tuning until there's a solution to the dynamic tuning problem, because the dynamic tuning is not great in my opinion and it's not something you want to force on users.

olysonek:
Actually I think there is a solution to the dynamic tuning problem, but it'll require a specific way of merging the pre/post-loaded profile with the other active profiles.  The reapply_sysctl problem should be solvable as well. I'll elaborate later.

defaults/main.yml
```diff
@@ -0,0 +1,67 @@
...
+# This is a list of the settings to apply using `sysctl`. Each list item is a
+# `dict` with allowed key names of `name`, `value`, `state`, and `previous`.
+# The settings are specified with a `name`, a `value`, and an optional `state`
+# or `previous` which controls the behavior.  For example:
+# kernel_settings_sysctl:
+#   - name: fs.epoll.max_user_watches
+#     value: 785592
+#   - name: fs.file-max
+#     value: 379724
+kernel_settings_sysctl: null
```
pcahyna, line 13: Wouldn't it be more natural to use an empty list as defaults for all those settings? Those settings are essentially lists of things to be set on the target system. By default there is nothing to be set if the user does not specify it -> this is expressed by an empty list.

richm: thanks - https://github.com/linux-system-roles/kernel_settings/pull/1/commits/b7cccb8db5289d08b8d9b03ca749abc10e5308d1

richm:
> I'm not saying that this is necessarily a deal breaker, but it is something that you should be aware of and something that you might want to document.

@olysonek https://github.com/linux-system-roles/kernel_settings/pull/1/commits/d474a0b8e13b88260607a8c21326a1613f354852

olysonek:
So I thought I was onto something with solving the `reapply_sysctl` problem, but I realized now that it's probably a no-go. Still, let me give you an outline of what I was thinking just in case it sparks any new ideas. Let me know if anything is unclear. (By the way, forget the different merge strategy that I mentioned above - I'm not proposing that here.)

I was thinking that Tuned could implement a way to override the `reapply_sysctl` setting directly in the profile for specific units. So in the `kernel_settings` profile, you could do something like
```
[sysctl]
reapply_sysctl=0
vm.swappiness=10
```
This way, the system sysctl settings would not be applied after applying the `[sysctl]` unit. However, this alone is not a good solution, because that would be almost equivalent to changing the `reapply_sysctl` setting in `/etc/tuned/tuned-main.conf` - it would disable applying the system sysctl settings completely, even for other sysctl units. So you would need a way to make the sysctl units somewhat separate. Luckily, you can already do that - you can have multiple units of the same type/plugin.

Let's say that the user has applied a profile `some_profile` with the following contents.
```
[sysctl]
vm.swappiness=20
```
Then the user tells the `kernel_settings` role to apply `vm.swappiness=10`. The `kernel_settings` role would write out the following `kernel_settings` profile:
```
[sysctl_kernel_settings]
type=sysctl
reapply_sysctl=0
vm.swappiness=10
```
If these two profiles are then applied together, the following would happen, in order:
1. `vm.swappiness=20` from the `[sysctl]` unit is applied
2. system sysctl settings are applied
3. `vm.swappiness=10` from the `[sysctl_kernel_settings]` unit is applied
4. and that's it.

Which is great! That's exactly what we need (in my opinion). But then I realized that verification is done for each unit separately. Which means that given that there are conflicting settings in the units, `tuned-adm verify` would always fail. Which is bad.

And that, kids, is the story of how olysonek spent half a day thinking about something which turned out to be bollocks ;).

So in conclusion, I don't see a way forward with this and documenting the problem is probably the best solution.

richm:
@olysonek ok - thanks for trying - so is the documentation in https://github.com/linux-system-roles/kernel_settings/pull/1/commits/d474a0b8e13b88260607a8c21326a1613f354852 sufficient?  If not, how should I change it?

olysonek:
> @olysonek ok - thanks for trying - so is the documentation in [d474a0b](https://github.com/linux-system-roles/kernel_settings/commit/d474a0b8e13b88260607a8c21326a1613f354852) sufficient? If not, how should I change it?

Yes it is, thanks. You have my lgtm for this PR :).

richm:
If there are no further comments, I would like to merge this PR

pcahyna:
Ok - I have some concerns about the interface but let's address them later.
(It is about the scalar values, each one of those has two variables, `something` and `something_state`, like:
```yaml
kernel_settings_transparent_hugepages: null
# set to `absent` to remove the setting
# kernel_settings_transparent_hugepages
kernel_settings_transparent_hugepages_state: null
```
I think there should be a better way to do it. But I don't want to hold the PR from being merged, so let's solve it in a separate issue.)

README.md
```diff
@@ -0,0 +1,280 @@
...
+`kernel_settings` role will cause this file to be reloaded and reapplied.  If
+this behavior is not desired, you will need to edit the `tuned` configuration
+on the managed hosts in `/etc/tuned/tuned-main.conf` and set
+`reapply_sysctl=0`.
+
+The settings you apply with the `kernel_settings` role may conflict with other
+settings.  For example, if you manually run the `sysctl` command, or manually
+edit `/etc/sysctl.d/` files, or if the `sysctl.d` files are installed by some
+system package, they may set the same values you are setting with the
+`kernel_settings` role.
```
pcahyna, line 270: @richm Can you please document which one takes precedence in case of conflict - the `sysctl.d` files or the settings done by the role? Or is it intentionally left unspecified?

richm: > @richm Can you please document which one takes precedence in case of conflict - the `sysctl.d` files or the settings done by the role? Or is it intentionally left unspecified?

https://github.com/linux-system-roles/kernel_settings/pull/5

PR #3, opened 2020-05-14
module should indicate if reboot needed to apply changes

richm:
Adds a new return value from the module `reboot_required` - this is
a boolean value, `false` by default.  If `true`, this means there
was a change made that will require rebooting the managed host in
order for the change to take effect.
I added a `RETURN` section to document the module return values.
I cleaned up the formatting of the examples and of the return
messages.

### DIFF
--- a/library/kernel_settings.py
+++ b/library/kernel_settings.py
@@ -210,7 +210,7 @@ kernel_settings:
 """
 
 RETURN = """
-message:
+msg:
     A short text message to say what action this module performed.
 new_profile:
     This is the tuned profile in dict format, after the changes, if any,
--- a/library/kernel_settings.py
+++ b/library/kernel_settings.py
@@ -218,6 +218,9 @@ new_profile:
 reboot_required:
     boolean - default is false - if true, this means a reboot of the
     managed host is required in order for the changes to take effect.
+active_profile:
+    This is the space delimited list of active profiles as reported
+    by tuned.
 """
 
 TUNED_PROFILE = os.environ.get("TEST_PROFILE", "kernel_settings")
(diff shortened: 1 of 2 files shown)


richm:
NOTE that there will be followup PR to use this functionality at the role level

library/kernel_settings.py
```diff
@@ -171,42 +171,53 @@ def wrapper_set_logger_class(_clsname):
...
+          state: absent
+"""
+
+RETURN = """
+message:
+    A short text message to say what action this module performed.
+new_profile:
+    This is the tuned profile in dict format, after the changes, if any,
+    have been applied.
+reboot_needed:
```
pcahyna, line 218: Please name it `reboot_required` for consistency with prior art here: https://docs.ansible.com/ansible/latest/modules/selinux_module.html#return-reboot_required

library/kernel_settings.py
```diff
@@ -171,42 +171,53 @@ def wrapper_set_logger_class(_clsname):
...
+          value: off
+        - name: nopti
+        - name: quiet
+          state: absent
+        - name: splash
+          state: absent
+"""
+
+RETURN = """
+message:
```
pcahyna, line 213: I believe that by convention this is called `msg`. https://docs.ansible.com/ansible/latest/reference_appendices/common_return_values.html#msg . Unless you want to keep `msg` only for error messages, but I do not see any prior art for this in existing core modules.

richm: fixed in https://github.com/linux-system-roles/kernel_settings/pull/3/commits/3d6c2d7828fa164c6fe89f5f6fdc3066218dd289

richm: sorry, fixed in https://github.com/linux-system-roles/kernel_settings/commit/3d6c2d7828fa164c6fe89f5f6fdc3066218dd289 and https://github.com/linux-system-roles/kernel_settings/pull/3/commits/20646a90b2043059fe7a890da8022459384b2160

richm:
please review - this PR along with its successor https://github.com/linux-system-roles/kernel_settings/pull/4 are required to unblock @RHEmployee PR https://github.com/linux-system-roles/kernel_settings/pull/7

richm:
Any more comments?  I'd like to merge this

richm:
> Any more comments? I'd like to merge this

If there are no further comments, I will merge this tomorrow

PR #4, opened 2020-05-15
handle reboot_needed; make changes take effect immediately

richm:
Added one new input parameter and one output variable.
`kernel_settings_reboot_ok` - used to tell the role if it is ok for the
role to reboot the machine if the changes need a reboot to be applied (e.g.
bootloader cmdline settings).

The role will export `kernel_settings_reboot_needed: true` if the user did
not set `kernel_settings_reboot_ok: true` and the changes need a reboot to
be applied.  The user is responsible for rebooting the machine when
convenient.

This also changes the handler behavior - now the handler only does the
reboot - the kernel_settings changes are now applied immediately, by
restarting tuned.  This makes the kernel_settings role more like the
`sysctl` module, which applies the changes immediately.  If the
kernel_settings module is called from other roles e.g. a role that deploys 
Elasticsearch, the changes have to be applied immediately, rather than
waiting until the end of the play containing the roles, when the handler is
executed, or Elasticsearch may not be able to start.  I also didn't want to
make it a requirement to force the user to use `meta: flush_handlers` after
calling the role.

This change also uses the role parameters explicitly in the test, to make
it more obvious to test writers what parameters there are and how they are
specified.


richm:
This is a follow on to https://github.com/linux-system-roles/kernel_settings/pull/3 - I wanted to submit this PR now, before it merges, because it is likely controversial and will require some review iterations.
@olysonek @yarda

tyll:
Thinking about this and the need for the network role to allow NM restarts - do we really need a "allow_reboot" variable for each role? Would someone allow reboot for kernel settings but not for selinux? Maybe this could be a generic system role flag that all roles use. If we need fine granularity, there could be at least a system-role default value that each role variable could use.

richm:
> Thinking about this and the need for the network role to allow NM restarts - do we really need a "allow_reboot" variable for each role?

Probably not.

> Would someone allow reboot for kernel settings but not for selinux?

Probably not.  They would probably want to have some sort of logic like this - at the end of the play, if any of the roles specified reboot needed, then reboot.

> Maybe this could be a generic system role flag that all roles use. If we need fine granularity, there could be at least a system-role default value that each role variable could use.

Sure, something like `linux_system_roles_reboot_ok` - once such a variable is introduced, I can change `kernel_settings/defaults/main.yml` - `kernel_settings_reboot_ok: "{{ linux_system_roles_reboot_ok | d(false) }}"` or something like that.

And also, the role would set `linux_system_roles_reboot_needed: true` instead of `kernel_settings_reboot_needed: true`

> 
> Regarding the network role, it could also use this info to restart NM when reboot is allowed.

richm:
@pcahyna re: https://github.com/linux-system-roles/kernel_settings/pull/3#pullrequestreview-413924222 - should I also change `kernel_settings_reboot_needed` to `kernel_settings_reboot_required`?

pcahyna:
> @pcahyna re: [#3 (review)](https://github.com/linux-system-roles/kernel_settings/pull/3#pullrequestreview-413924222) - should I also change `kernel_settings_reboot_needed` to `kernel_settings_reboot_required`?

yes please - prior art: https://github.com/linux-system-roles/selinux#selinux_reboot_required

defaults/main.yml
```diff
@@ -65,3 +65,17 @@ kernel_settings_bootloader_cmdline: []
...
+# applied, and will set kernel_settings_reboot_needed: true
+kernel_settings_reboot_ok: false
+
+# This is a return value from the role. If true, the role applied some change
+# which will require the managed host to be rebooted in order for the change
+# to take effect.  You can use this value to see if a reboot is needed, and
+# then reboot the machine at a covenient time in your playbook. If you would
+# rather have the role reboot the managed host when necessary, see
+# kernel_settings_reboot_ok
+kernel_settings_reboot_needed: false
```
pcahyna, line 81: I think it should not be in defaults - better make it a fact. Defaults are for things that the user may want to override, but this is not meant to be overriden.

tests/tests_simple_settings.yml
```diff
@@ -1,31 +1,62 @@
...
+        kernel_settings_transparent_hugepages_defrag: "{{
+          kernel_settings_test_thd | d(omit)
+        }}"
+        kernel_settings_transparent_hugepages_defrag_state: "{{
+          kernel_settings_test_thd_state | d(omit)
+        }}"
+        kernel_settings_purge: "{{ kernel_settings_test_purge | d(false) }}"
+        kernel_settings_reboot_ok: "{{
+          kernel_settings_test_reboot_ok | d(false)
+        }}"
```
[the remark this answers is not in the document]
pcahyna: I also feel that the change is quite unrelated to the topic of the commit or even of the PR.

richm:
@yarda @olysonek @RHEmployee I had to do https://github.com/linux-system-roles/kernel_settings/pull/4/commits/f7fa956e93f576096946413502519b6bfb0b3983 to make it so that the bootloader cmdline written by `tuned` works correctly

richm:
I filed a tuned issue about the bootloader cmdline https://github.com/redhat-performance/tuned/issues/268

It seems the problem is only with Fedora Cloud image based VMs.  The issue does not seem to affect Fedora running on bare metal, or Fedora VMs not based on the cloud image.  Unfortunately, this affects our primary test platform (local VMs using the Fedora Cloud image), and other important platforms (OpenStack VMs using the Fedora Cloud image).  Since this doesn't affect all Fedora users, and only affects bootloader command line settings, I would like to keep support for Fedora, including CI, and I would like to do that by using the workaround in https://github.com/linux-system-roles/kernel_settings/commit/f7fa956e93f576096946413502519b6bfb0b3983

For end users, there is a workaround - use `grubby` to add a bootloader cmdline argument e.g. `grubby --update-kernel=ALL --args=quiet` - this is what the CI test does.

pcahyna:
> I filed a tuned issue about the bootloader cmdline [redhat-performance/tuned#268](https://github.com/redhat-performance/tuned/issues/268)

:+1: 

>  Since this doesn't affect all Fedora users, and only affects bootloader command line settings, I would like to keep support for Fedora, including CI, and I would like to do that by using the workaround in [f7fa956](https://github.com/linux-system-roles/kernel_settings/commit/f7fa956e93f576096946413502519b6bfb0b3983)

Do you mean that you want to move the workaround from the test playbooks to the role task file? This is IMO the right thing to do, because those images can actually be used in production and should work out of the box.

richm:
> > I filed a tuned issue about the bootloader cmdline [redhat-performance/tuned#268](https://github.com/redhat-performance/tuned/issues/268)
> 
> +1
> 
> > Since this doesn't affect all Fedora users, and only affects bootloader command line settings, I would like to keep support for Fedora, including CI, and I would like to do that by using the workaround in [f7fa956](https://github.com/linux-system-roles/kernel_settings/commit/f7fa956e93f576096946413502519b6bfb0b3983)
> 
> Do you mean that you want to move the workaround from the test playbooks to the role task file? This is IMO the right thing to do, because those images can actually be used in production and should work out of the box.

The latest commit does not have any special bootloader cmdline handling in the main role code.  For testing, the `tests/tests_simple_settings.yml` test is done in such a way that the bootloader cmdline settings are tested on RedHat and CentOS but not on Fedora.  This means that the role should work on all supported platforms, including Fedora if the user is using an updated image - the role nor the test tries to verify the correct version of the bootloader.  It is pretty difficult to try to determine and workaround the buggy bootloader in Fedora.

I would like to get this PR merged to unblock @RHEmployee testing e.g. https://github.com/linux-system-roles/kernel_settings/pull/7 - then I will have another PR to address the bootloader issues e.g. documentation.

handlers/main.yml
```diff
@@ -1,11 +1,18 @@
 ---
-- name: restart kernel_settings services
-  service:
-    name: "{{ item }}"
-    state: restarted
-  loop: "{{ __kernel_settings_services }}"
```
pcahyna, line 6: What if there have been changes, but they do not require a reboot? Shouldn't you keep the "restart" task in this case?

richm: > What if there have been changes, but they do not require a reboot? Shouldn't you keep the "restart" task in this case?

It wasn't necessary to restart the service in order to apply the changes, all I need to do is call `tuned-adm profile ACTIVE_PROFILE` - https://github.com/linux-system-roles/kernel_settings/pull/4/files#diff-2444ad0870f91f17ca6c2a5e96b26823R93

And - in order to match the semantics of similar modules such as `sysctl`, the changes are not applied in the handler, they are applied immediately - this meets the expectations of tests such as https://github.com/linux-system-roles/kernel_settings/pull/7/files#diff-0363d71adca523f53e6765d01762b5c0R44 that expect the values are applied immediately after calling the role.

pcahyna: Ah, I see this is explained in the commit message, sorry.

tests/tests_simple_settings.yml
```diff
@@ -1,31 +1,50 @@
...
+      include_vars:
+        file: "{{ item }}"
+      loop: "{{
+        q('first_found', __kernel_settings_test_vars_files, errors='ignore')
+      }}"
+      vars:
+        __kernel_settings_test_vars_files:
+          - "vars/vars_simple_settings_{{ ansible_distribution }}_\
+            {{ ansible_distribution_major_version }}.yml"
+          - "vars/vars_simple_settings_{{ ansible_distribution }}.yml"
```
pcahyna, line 19: I feel this is overcomplicated for the task (skipping a test on Fedora 31). A single `when` condition on `include_role` should be enough, like here: https://github.com/linux-system-roles/network/blob/master/tests/tests_states_nm.yml#L20 . (Fedora 32 works fine, by the way?)

pcahyna:
I agree with not having any special bootloader cmdline handling in the main role code - skipping the test on Fedora 31 could be probably done in a simpler way, though.

tests/tests_simple_settings.yml
richm: I don't want to skip the entire test on Fedora (note that there is a bootloader cmdline problem on both Fedora 31 and Fedora 32 cloud images) - I want to test everything *except* the bootloader cmdline settings on Fedora - I suppose I could revert back to having all of the test settings in the single `vars/vars_simple_settings.yml`, and do something like this?
```yaml
- name: disable bootloader cmdline testing on Fedora
  set_fact:
    kernel_settings_bootloader_cmdline: none
    __kernel_settings_blcmdline_value: none
    __kernel_settings_check_reboot: false
  when: ansible_distribution == "Fedora"
```
?

tests/tests_simple_settings.yml
```diff
@@ -6,17 +6,12 @@
...
-        __kernel_settings_test_vars_files:
-          - "vars/vars_simple_settings_{{ ansible_distribution }}_\
-            {{ ansible_distribution_major_version }}.yml"
-          - "vars/vars_simple_settings_{{ ansible_distribution }}.yml"
+    - name: disable bootloader cmdline testing on Fedora
+      set_fact:
+        kernel_settings_bootloader_cmdline: null
+        __kernel_settings_blcmdline_value: null
+        __kernel_settings_check_reboot: false
+      when: ansible_distribution == "Fedora"
```
richm, line 14: @pcahyna is this better?

pcahyna: yes, I believe this is better, thanks.

tests/tests_simple_settings.yml
pcahyna: Sorry, I did not understand that the error persists on Fedora 32 as well - is it fixed in rawhide, or what is your strategy for supporting Fedora? The problem should not remain unfixed forever.

richm: > Sorry, I did not understand that the error persists on Fedora 32 as well - is it fixed in rawhide, or what is your strategy for supporting Fedora? The problem should not remain unfixed forever.

I will have a follow-on PR which is for the documentation, explaining that the Fedora Cloud images for F31 and F32 are problematic:
* Fedora uses BootLoaderSpec config snippets in /boot/loader/config/, but the initial /boot/loader/config/$machineid-$version.conf file machineid does not match the /etc/machine-id in the image.  This causes the kernel_settings role (via tuned) not to set the bootloader cmdline properly.  This is fixed by doing a `dnf update` of the `grub*` packages.
* Fedora builds the cloud images with a buggy bootloader that does not add a space after user settable bootloader variables.  This causes the kernel_settings role (via tuned) not to set the bootloader cmdline properly.  This is fixed by doing a `dnf update` of the `grub*` packages and rebooting.

My hope is that
* Fedora will rebuild and publish newer cloud images - I am trying to find out where to request that to be done
* Fedora 33 will finally fix the bootloader in the cloud images

pcahyna: @richm I am especially interested in the last point - is rawhide fixed already? If not, what fixes does it need?

richm: > @richm I am especially interested in the last point - is rawhide fixed already? If not, what fixes does it need?

I tried http://mirrors.kernel.org/fedora/development/rawhide/Cloud/x86_64/images/Fedora-Cloud-Base-Rawhide-20200603.n.0.x86_64.qcow2 - this is even more broken than the f31 and f32 images - none of the workarounds help.

PR #5, opened 2020-05-15
clarify precedence of settings w.r.t. sysctl, sysfs, bootloader

richm:
Clarify the precedence of setting parameters with the `kernel_settings`
role vs. using `sysctl` files, the `sysctl` command, `grubby`, etc.


richm:
@olysonek

PR #6, opened 2020-05-15
improve testing docs

richm:
improve testing docs, especially to help first time test writers

PR #7, opened 2020-05-18
tests_change_settings: Add reboot/change test.

RHEmployee:
Should test if settings are applied after reboot.


tests/tests_change_settings.yml
```diff
@@ -0,0 +1,78 @@
...
+            state: absent
+        kernel_settings_sysfs:
+          - name: /sys/kernel/debug/x86/nmi_longest_ns
+            value: 2000000
+        kernel_settings_bootloader_cmdline:
+          - name: spectre_v2
+            value: "on"
+          - name: quiet
+            state: absent
+
```
richm, line 43: Unless this test only cares about checking to see if the settings are applied after reboot, it should check right after the role is run to see if the kernel settings are applied to the system.

tests/tests_change_settings.yml
```diff
@@ -0,0 +1,78 @@
...
+    - name: reboot the machine - see if settings persist after reboot
+      reboot:
+        test_command: tuned-adm active
+
+    - name: check after reboot for ns_longest
+      shell: 'grep 2000000 /sys/kernel/debug/x86/nmi_longest_ns'
```
richm, line 49: `command: grep 2000000 /sys/kernel/debug/x86/nmi_longest_ns`
We avoid using `shell` if possible - and you don't need the single quotes around the string

tests/tests_change_settings.yml
```diff
@@ -0,0 +1,78 @@
...
+      
+    - name: check after reboot cmdline
+      shell: 'grep spectre_v2 /proc/cmdline'
```
richm, line 52: `command: grep spectre_v2 /proc/cmdline`

tests/tests_change_settings.yml
```diff
@@ -0,0 +1,78 @@
...
+            value: "off"
+          - name: quiet
+            state: absent
+
+    - name: reboot the machine - see if settings persist after reboot
+      reboot:
+        test_command: tuned-adm active
+
+    - name: check after reboot cmdline
+      shell: 'grep spectre_v2 /proc/cmdline'
```
richm, line 77: `command: grep spectre_v2=off /proc/cmdline`

tests/tests_change_settings.yml
```diff
@@ -0,0 +1,78 @@
...
+
```
nhosoi, line 78: ansible-lint does not like the blank lines at the end of the files.

tests/tests_change_settings.yml
RHEmployee: Hi. Thanks. So do you want me to add more checks right after settings were applied?

tests/tests_change_settings.yml
RHEmployee: Hello. Thanks. So what I should use instead?

RHEmployee:
Hi. During testing I just found:
TASK [reboot the machine - see if settings persist after reboot] ***********************************************************************
task path: /home/rhack/system-role/devel/kernel_settings/tests/tests_change_settings.yml:44
changed: [10.0.138.119] => {"changed": true, "elapsed": 39, "rebooted": true}

TASK [check after reboot for ns_longest] ***********************************************************************************************
task path: /home/rhack/system-role/devel/kernel_settings/tests/tests_change_settings.yml:48
changed: [10.0.138.119] => {"changed": true, "cmd": "grep 2000000 /sys/kernel/debug/x86/nmi_longest_ns", "delta": "0:00:00.007257", "end": "2020-05-19 04:00:02.068855", "rc": 0, "start": "2020-05-19 04:00:02.061598", "stderr": "", "stderr_lines": [], "stdout": "2000000", "stdout_lines": ["2000000"]}

TASK [check after reboot cmdline] ******************************************************************************************************
task path: /home/rhack/system-role/devel/kernel_settings/tests/tests_change_settings.yml:51
fatal: [10.0.138.119]: FAILED! => {"changed": true, "cmd": "grep spectre_v2 /proc/cmdline", "delta": "0:00:00.006748", "end": "2020-05-19 04:00:03.209959", "msg": "non-zero return code", "rc": 1, "start": "2020-05-19 04:00:03.203211", "stderr": "", "stderr_lines": [], "stdout": "", "stdout_lines": []}

RUNNING HANDLER [linux-system-roles.kernel_settings : restart kernel_settings services] ************************************************
task path: /home/rhack/system-role/devel/kernel_settings/handlers/main.yml:2

RUNNING HANDLER [linux-system-roles.kernel_settings : verify settings] *****************************************************************
task path: /home/rhack/system-role/devel/kernel_settings/handlers/main.yml:9

Would you please investigate it? Thank you very much.

tests/tests_change_settings.yml
RHEmployee: Thanks. Fixed in another commit.

tests/tests_change_settings.yml
RHEmployee: Thank you. Removed in another commit.

tests/tests_change_settings.yml
RHEmployee: Thank you. Removed in another commit.

tests/tests_change_settings.yml
RHEmployee: Also: Removed ' chars in another commit.

tests/tests_change_settings.yml
richm: Yes.

tests/tests_change_settings.yml
```diff
@@ -0,0 +1,77 @@
...
+            value: "on"
+          - name: quiet
+            state: absent
+
+    - name: reboot the machine - see if settings persist after reboot
+      reboot:
+        test_command: tuned-adm active
+
+    - name: check after reboot for ns_longest
+      shell: grep 2000000 /sys/kernel/debug/x86/nmi_longest_ns
```
richm, line 49: I think `tuned-adm verify` checks to see if *all* of the settings in the profile are correctly applied to the system. @yarda @olysonek please confirm

richm: Also note - use `command` instead of `shell` unless you need to use bash logic, piping, or variable constructs

tests/tests_change_settings.yml
```diff
@@ -0,0 +1,77 @@
...
+      
+    - name: check after reboot cmdline
+      shell: grep spectre_v2 /proc/cmdline
```
richm, line 52: I think `tuned-adm verify -i` checks this too.

richm:
@yarda @olysonek re: https://github.com/linux-system-roles/kernel_settings/pull/7#issuecomment-630654598

This looks like a bug in tuned on Fedora 31.  Steps:

* start and enable tuned
* configure the profile kernel_settings/tuned.conf with bootloader cmdline like this:
```
[bootloader]
cmdline = spectre_v2=on nopti splash
```
* apply profile `tuned-adm profile kernel_settings`
* check `/etc/tuned/bootcmdline` - it contains this:
```
TUNED_BOOT_CMDLINE="spectre_v2=on nopti splash"
TUNED_BOOT_INITRD_ADD=""
```
* check `/etc/grub*` and `/etc/grub*/*` and `/etc/default/grub` - looks like all of the tuned modifications are there?
* reboot
* check `/proc/cmdline` - FAILS - does not have `spectre_v2=on nopti splash`

I'm not sure what's going on - I would appreciate any help

tests/tests_change_settings.yml
```diff
@@ -0,0 +1,77 @@
...
+    - name: check after reboot tuned
+      shell: tuned-adm verify -i
+
+    - name: apply kernel_settings again
+      include_role:
+        name: linux-system-roles.kernel_settings
+      vars:
+        kernel_settings_sysctl:
+          - name: fs.file-max
+            value: absent
```
richm, line 63: Did you mean `state: absent` i.e. you want to remove the `fs.file-max` value?  Otherwise, `absent` is not a valid value.

tests/tests_change_settings.yml
```diff
@@ -0,0 +1,77 @@
...
+            value: "off"
+          - name: quiet
+            state: absent
+
+    - name: reboot the machine - see if settings persist after reboot
+      reboot:
+        test_command: tuned-adm active
+
+    - name: check after reboot cmdline
+      shell: grep spectre_v2 /proc/cmdline
```
richm, line 77: use `command` instead of `shell` - use `spectre_v2=off` instead of just `spectre_v2`

tests/tests_change_settings.yml
richm: > Hello. Thanks. So what I should use instead?

Use the `command` module instead of `shell`:
```
command: grep 2000000 /sys/kernel/debug/x86/nmi_longest_ns
```
But see my other comment that you can use `tuned-adm verify -i` to verify all settings

tests/tests_change_settings.yml
RHEmployee: Thanks. Added in another commit.

tests/tests_change_settings.yml
RHEmployee: Thanks. Changed to command everywhere.
It's much easier to handle test results when I see precisely where test failed.

tests/tests_change_settings.yml
RHEmployee: Ok. I switched to command everywhere but I need pipes. Ok. Need to review again and change.

tests/tests_change_settings.yml
RHEmployee: Hello. It's not much clear here.
I used:
tuned/etc/tuned/basic_settings/tuned.conf
which contains:
fs.file-max=379724
so i want to reset it to default I think.

tests/tests_change_settings.yml
RHEmployee: Thanks. Fixed.

RHEmployee:
Hello. Another interesting issue when new commits are used:
changed: [10.0.136.13] => {"changed": true, "message": "Kernel settings were updated", "new_profile": {"bootloader": {"cmdline": "spectre_v2=on"}, "sysctl": {"fs.file-max": "400000"}, "sysfs": {"/sys/class/net/lo/mtu": "65000"}}, "original_message": null}

TASK [check sysfs] *********************************************************************************************************************
task path: /home/rhack/system-role/devel/kernel_settings/tests/tests_change_settings.yml:44
fatal: [10.0.136.13]: FAILED! => {"changed": true, "cmd": ["grep", "65000", "/sys/class/net/lo/mtu"], "delta": "0:00:00.004921", "end": "2020-05-20 05:21:36.445608", "msg": "non-zero return code", "rc": 1, "start": "2020-05-20 05:21:36.440687", "stderr": "", "stderr_lines": [], "stdout": "", "stdout_lines": []}

RHEmployee:
> Hello. Another interesting issue when new commits are used:
> changed: [10.0.136.13] => {"changed": true, "message": "Kernel settings were updated", "new_profile": {"bootloader": {"cmdline": "spectre_v2=on"}, "sysctl": {"fs.file-max": "400000"}, "sysfs": {"/sys/class/net/lo/mtu": "65000"}}, "original_message": null}
> 
> TASK [check sysfs] *********************************************************************************************************************
> task path: /home/rhack/system-role/devel/kernel_settings/tests/tests_change_settings.yml:44
> fatal: [10.0.136.13]: FAILED! => {"changed": true, "cmd": ["grep", "65000", "/sys/class/net/lo/mtu"], "delta": "0:00:00.004921", "end": "2020-05-20 05:21:36.445608", "msg": "non-zero return code", "rc": 1, "start": "2020-05-20 05:21:36.440687", "stderr": "", "stderr_lines": [], "stdout": "", "stdout_lines": []}

Hello. Looks like role is not able to set this sysfs value. However, test ends when first error is reached so, if I need to continue, I need to modify test itself. Or is here some other way?

richm:
> Hello. Looks like role is not able to set this sysfs value.

Can you provide the entire test log?

> However, test ends when first error is reached so, if I need to continue, I need to modify test itself. Or is here some other way?

The `tests/tests_simple_settings.yml` test for kernel_settings is a "fail last" test.  That is, when the test has validation errors, it simply note that an error has occurred, and will continue until the end where it will report that an error has occurred somewhere.

https://github.com/linux-system-roles/kernel_settings/blob/master/tests/tasks/assert_kernel_settings.yml#L9

This will set the variable `__kernel_settings_success: true` at the beginning of the test - the test assumes success at first.

https://github.com/linux-system-roles/kernel_settings/blob/master/tests/tasks/assert_kernel_settings_conf_files.yml#L18

This test task runs with `ignore_errors: true` so that the task will not cause Ansible to immediately abort upon failure.  It also stores the result in the variable `__kernel_settings_register_profile_conf_result`.

https://github.com/linux-system-roles/kernel_settings/blob/master/tests/tasks/assert_kernel_settings_conf_files.yml#L31

This task will set `__kernel_settings_success: false` if the previous task failed.  Other test tasks work in a similar fashion - execute the test task with `ignore_errors: true` and use `register:` to save the result in a variable, then have a separate task to set `__kernel_settings_success: false` if the test task failed.

At the end, report that an error occurred:  https://github.com/linux-system-roles/kernel_settings/blob/master/tests/tasks/assert_kernel_settings.yml#L16

The user will have to scroll back through the log to find out where the actual test failures happened.

RHEmployee:
1) log is here:
https://pastebin.com/8vaKBmPb

2) ok. thanks. Need to add to my code and test.

richm:
>     1. log is here:
>        https://pastebin.com/8vaKBmPb

This will be fixed by https://github.com/linux-system-roles/kernel_settings/pull/4

I might need to split that PR into two parts, and fix the bootloader stuff separately.

> 
>     2. ok. thanks. Need to add to my code and test.

richm:
@RHEmployee please see https://github.com/linux-system-roles/kernel_settings/pull/6/files and especially https://github.com/linux-system-roles/kernel_settings/pull/6/commits/fa71b5e9140fa2889b1336a0932b98ab105d75e6

RHEmployee:
Hi. Thanks. I added:
ignore_errors: true
to my cases.

tests/tests_change_settings.yml
RHEmployee: Should we rely on tuned backend?

tests/tests_change_settings.yml
RHEmployee: What I should use to remove(?) value from profile?

tests/tests_change_settings.yml
richm: > Should we rely on tuned backend?

For now, I would say yes, until such time as we support a different provider (we don't call them backends, we call them providers).  However, if you feel that we should explicitly call `sysctl` to check `sysctl` parameters, look at files under `/sys` to check `sysfs` parameters, etc. then that's fine.

I don't think we have general guidelines for provider specific testing - @pcahyna please correct me if I'm wrong.  For example - the network role has two providers - NetworkManager (`nm`) and SysV Init scripts (`initscripts`) - https://github.com/linux-system-roles/network/tree/master/tests - As you can see here there are tests like `tests_bridge_nm.yml` and `tests_bridge_initscripts.yml` - these set the `network_provider: nm` or `network_provider: initscripts` then run the role and verify the results.  The code that does the verification seems mostly to be written to handle both providers.  For example - https://github.com/linux-system-roles/network/blob/master/tests/tasks/get_profile_stat.yml - handles both the `nm` case and the `initscripts` case.

tests/tests_change_settings.yml
richm: > Ok. I switched to command everywhere but I need pipes. Ok. Need to review again and change.

I don't think you need pipes or redirection here.  But I suppose since this is test code and not run-time code, we can relax some of the restrictions against using `shell` and other unsafe constructs, in the interest of making it easier for test writers.

tests/tests_change_settings.yml
richm: > What I should use to remove(?) value from profile?

Use `state: absent`.  If you have an existing tuned profile with
```
[sysctl]
fs.file-max=379724
```
and you want to remove that setting, use
```yaml
kernel_settings_sysctl:
  - name: fs.file-max
    state: absent
```
This will remove the setting from the tuned profile.  You can check the tuned profile file to see that the settings have been removed.

However, It will *not* change the system to revert the value back to its original system default value.  But this is the same as with tuned - how do you test this in tuned?  That is, when you are testing tuned, and you update a profile to remove a `sysctl` setting, and you reapply the profile, how do you verify that the `sysctl` setting was removed?

One way to check would be to rely on the fact that a system reboot is supposed to restore the value to its original setting - query the system to get the default value, save it, apply the profile to add the setting, run the role to remove the setting, reboot, query the value, verify that the value is the same as the original value
* start the test with a clean system
* `command: sysctl ...` to get the default value from the system
* apply the basic settings tuned profile - set the value to a different value
* `command: sysctl ...` or `tuned-adm verify -i` to verify that the setting is applied correctly
* run the kernel_settings role to remove the value using `state: absent` or `state: empty`
* reboot
* `command: sysctl ...` to get the value from the system - verify that this value after reboot is the same as the original starting value

tests/tests_change_settings.yml
```diff
@@ -0,0 +1,109 @@
...
+        src: tuned/etc/tuned/basic_settings/tuned.conf
+        dest: /etc/tuned/kernel_settings/tuned.conf
+        mode: 0644
+
+    - name: Ensure required services are enabled and started
+      service:
+        name: tuned
+        state: restarted
+        enabled: yes
+
```
richm, line 25: Sorry - forgot that there is a missing step here - to apply the kernel_settings profile:
```yaml
    - name: apply kernel_settings tuned profile
      command: tuned-adm profile kernel_settings
```

You might also want to add a check here to make sure your settings have been applied correctly e.g. `tuned-adm verify -i` or checking individual `sysctl`, `sysfs`, etc. settings.  Note that if your profile sets `[bootloader]` `cmdline` settings, `tuned-adm verify -i` will report failure unless you first reboot the system.

tests/tests_change_settings.yml
richm: see https://github.com/linux-system-roles/kernel_settings/pull/7#discussion_r431920352

richm:
If you use `ignore_errors: true` for all of the tasks, then the test will always pass and never fail.

RHEmployee:
Hello. Yeah. It's true:
PLAY RECAP *****************************************************************************************************************************
10.0.137.31                : ok=34   changed=22   unreachable=0    failed=1    skipped=0    rescued=0    ignored=7

However:
How I should develop my test when there is bugs which are solved elsewhere? I need to add cases and not be hold by one bug... or one fail.
Would you please provide code snippet, which I just put after my cases to not ignore them but continue after fail? It's not much clear from previous comment about variables.

tests/tests_change_settings.yml
RHEmployee: Hello. Should be solved by last addition:
     vars:
        kernel_settings_sysctl:
          - name: fs.file-max
            value: absent
          - name: kernel.threads-max
            state: absent
          - name: spectre_v2
            state: absent
          - name: quiet
            state: absent

    - name: reboot the machine - see if settings persist after reboot
      reboot:
        test_command: tuned-adm active

    - name: check after spectre_v2 reboot
      command: grep -vq spectre_v2 /proc/cmdline

    - name: check sysctl after reboot
      shell: sysctl -n fs.file-max | grep -vq 400000

RHEmployee:
Note: I removed ignores in last commit.

tests/tests_change_settings.yml
RHEmployee: Ok. Added to test.
So kernel roles are now tuned testing.

tests/tests_change_settings.yml
RHEmployee: Hello. Thanks. So what do you propose to use to check runtime values?

pcahyna:
> Note: I removed ignores in last commit.

Did it work? Apparently yes. What was the problem?

tests/tests_change_settings.yml
pcahyna: @RHEmployee there is still `value: absent` there. This sets `fs.file-max` literally to the string "absent" instead of removing it. Does not sound right.

RHEmployee: Hello. Thank you. So what I should use instead?

RHEmployee:
> > Note: I removed ignores in last commit.
> 
> Did it work? Apparently yes. What was the problem?

So should I return ignore error back?

pcahyna:
> > > Note: I removed ignores in last commit.
> > 
> > 
> > Did it work? Apparently yes. What was the problem?
> 
> So should I return ignore error back?

No, I am asking why was it ever needed if the CI passes without it. But now I see that there are no runs of the test-harness CI, only of Teavis - so the CI is not running correctly?

richm:
> > > > Note: I removed ignores in last commit.
> > > 
> > > 
> > > Did it work? Apparently yes. What was the problem?
> > 
> > 
> > So should I return ignore error back?
> 
> No, I am asking why was it ever needed if the CI passes without it. But now I see that there are no runs of the test-harness CI, only of Travis - so the CI is not running correctly?

I added `needs-ci` label so that @RHEmployee can trigger CI for this PR.

richm:
@pcahyna what is the best practices recommendation for developing tests for system roles - to fail fast or fail last?  If it is to fail fast, then all of this coding around using `ignore_errors: true` and having a separate task to set an error flag is not needed.

richm:
> How I should develop my test when there is bugs which are solved elsewhere?

Ideally by fixing the bug that is blocking your testing - https://github.com/linux-system-roles/kernel_settings/pull/4 should be merged

> I need to add cases and not be hold by one bug... or one fail.

richm:
> Would you please provide code snippet, which I just put after my cases to not ignore them but continue after fail? It's not much clear from previous comment about variables.

```yaml
    - name: check cmdline for spectre_v2 after reboot
      command: grep spectre_v2=on /proc/cmdline
      ignore_errors: true
      register: kernel_settings_register_test

    - name: set error flag if cmdline check failed
      set_fact:
        kernel_settings_test_success: false
      when: kernel_settings_register_test is failed
```

The `ignore_errors: true` tells Ansible not to fail immediately.  The `register: kernel_settings_register_test` saves the task result.  The `set_fact` task will set the overall test success/failure flag if the previous task failed.

tests/tests_change_settings.yml
pcahyna: Depends on what you want to do, but if you want to remove the setting, use `state: absent`, as mentioned already in https://github.com/linux-system-roles/kernel_settings/pull/7/files#r431920352

tests/tests_change_settings.yml
```diff
@@ -0,0 +1,101 @@
...
+    - name: reboot the machine - see if settings persist after reboot
+      reboot:
+        test_command: tuned-adm active
+
+    - name: check after spectre_v2 reboot
+      command: grep -vq spectre_v2 /proc/cmdline
+
+    - name: check sysctl after reboot
+      shell: sysctl -n fs.file-max | grep -vq 400000
```
richm, line 101: you can also do `command: grep -vq 400000 /proc/sys/fs/file-max` - as you are probably aware, every setting known by `sysctl` is really a file under `/proc/sys`

pcahyna:
> @pcahyna what is the best practices recommendation for developing tests for system roles - to fail fast or fail last? If it is to fail fast, then all of this coding around using `ignore_errors: true` and having a separate task to set an error flag is not needed.

Our tests have been always failing fast. Coding them in a way that thy fail last would be really unwieldy. I suppose that the gap here is that we don;t really have a way to mark the test as xfail. We could use `ignore_errors:` for this (possibly with something like `ignore_errors: "{{ xfail }}` and set the `xfail` variable at the beginning of the playbook, if `ignore_errors` supports variable expansion), or skip the test using tags: `tags: [never, skip]`.

richm:
ok - so @RHEmployee - I would say don't worry about using `ignore_errors: true` and `register: kernel_settings_register_test` unless there is some reason you need to perform some checks after some other checks have failed.  And if you are stuck because the test fails on the first check, and you cannot make any good progress with test writing, it is because we're still waiting for https://github.com/linux-system-roles/kernel_settings/pull/4 to be merged

tests/tests_change_settings.yml
RHEmployee: Ok. Solved. Thanks.

richm:
@RHEmployee #4 was merged - please pull and rebase your code on top of the latest master

RHEmployee:
> @RHEmployee #4 was merged - please pull and rebase your code on top of the latest master

Hello. Great news! Thank you. Is here some documentation/manual, how to do it? Thank you very much.

richm:
> > @RHEmployee #4 was merged - please pull and rebase your code on top of the latest master
> 
> Hello. Great news! Thank you. Is here some documentation/manual, how to do it? Thank you very much.

Are you familiar with the `git rebase` command?  The following assumes you have a git remote called `origin` that points to https://github.com/linux-system-roles/kernel_settings.git, and a git remote called `rhack` that points to git@github.com:RHEmployee/linux-system-roles-kernel_settings.git - For example, in my local copy of the kernel_settings role I have the following git remotes:
```
$ git remote -v
origin	https://github.com/linux-system-roles/kernel_settings.git (fetch)
origin	https://github.com/linux-system-roles/kernel_settings.git (push)
richm	git@github.com:richm/linux-system-roles-kernel_settings.git (fetch)
richm	git@github.com:richm/linux-system-roles-kernel_settings.git (push)
```
the following also assumes you are doing your work on your local master branch, and you are currently on your master branch.  Run the following commands:
```
git commit -a -m "wip" # if you have unsaved changes on your local master branch
git fetch origin # pull the latest changes from https://github.com/linux-system-roles/kernel_settings.git
git rebase origin/master # rebase your changes on top of the latest origin master branch changes
```
I don't think you will have any rebase conflicts.  If you do, and are not sure what to do, let me know.

RHEmployee:
Hello. Finally, I was able to pass some tests after rebase and apply some tweaks.
However, I'm not able pass last part whit state: absent when quiet cmdline is used.
LOG:
TASK [check sysfs after reboot] ********************************************************************************************************
task path: /home/rhack/system-role/devel/kernel_settings/tests/tests_change_settings.yml:60
changed: [10.0.137.196] => {"changed": true, "cmd": ["grep", "65000", "/sys/class/net/lo/mtu"], "delta": "0:00:00.004478", "end": "2020-
06-05 07:42:38.255238", "rc": 0, "start": "2020-06-05 07:42:38.250760", "stderr": "", "stderr_lines": [], "stdout": "65000", "stdout_lin
es": ["65000"]}

TASK [check cmdline for spectre_v2 after reboot] ***************************************************************************************
task path: /home/rhack/system-role/devel/kernel_settings/tests/tests_change_settings.yml:63
changed: [10.0.137.196] => {"changed": true, "cmd": ["grep", "spectre_v2=on", "/proc/cmdline"], "delta": "0:00:00.004623", "end": "2020-
06-05 07:42:39.304789", "rc": 0, "start": "2020-06-05 07:42:39.300166", "stderr": "", "stderr_lines": [], "stdout": "BOOT_IMAGE=(hd0,msd
os1)/boot/vmlinuz-4.18.0-205.el8.x86_64 root=UUID=d10f71f9-35e2-4f38-9b1e-e0e8e62b08c7 ro console=tty0 console=ttyS0,115200 crashkernel=
auto net.ifnames=0 rhgb quiet spectre_v2=on nopti splash", "stdout_lines": ["BOOT_IMAGE=(hd0,msdos1)/boot/vmlinuz-4.18.0-205.el8.x86_64
root=UUID=d10f71f9-35e2-4f38-9b1e-e0e8e62b08c7 ro console=tty0 console=ttyS0,115200 crashkernel=auto net.ifnames=0 rhgb quiet spectre_v2
=on nopti splash"]}

TASK [show cmdline] ********************************************************************************************************************
task path: /home/rhack/system-role/devel/kernel_settings/tests/tests_change_settings.yml:66
changed: [10.0.137.196] => {"changed": true, "cmd": ["cat", "/proc/cmdline"], "delta": "0:00:00.004452", "end": "2020-06-05 07:42:40.380
167", "rc": 0, "start": "2020-06-05 07:42:40.375715", "stderr": "", "stderr_lines": [], "stdout": "BOOT_IMAGE=(hd0,msdos1)/boot/vmlinuz4.18.0-205.el8.x86_64 root=UUID=d10f71f9-35e2-4f38-9b1e-e0e8e62b08c7 ro console=tty0 console=ttyS0,115200 crashkernel=auto net.ifnames=0
 rhgb quiet spectre_v2=on nopti splash", "stdout_lines": ["BOOT_IMAGE=(hd0,msdos1)/boot/vmlinuz-4.18.0-205.el8.x86_64 root=UUID=d10f71f9
-35e2-4f38-9b1e-e0e8e62b08c7 ro console=tty0 console=ttyS0,115200 crashkernel=auto net.ifnames=0 rhgb quiet spectre_v2=on nopti splash"]
}

TASK [check cmdline for quiet flag after reboot (should not present)] ******************************************************************
task path: /home/rhack/system-role/devel/kernel_settings/tests/tests_change_settings.yml:69
fatal: [10.0.137.196]: FAILED! => {"changed": true, "cmd": ["grep", "-Lvq", "quiet", "/proc/cmdline"], "delta": "0:00:00.004199", "end":
 "2020-06-05 07:42:41.421215", "msg": "non-zero return code", "rc": 1, "start": "2020-06-05 07:42:41.417016", "stderr": "", "stderr_line
s": [], "stdout": "", "stdout_lines": []}

RUNNING HANDLER [linux-system-roles.kernel_settings : reboot the managed host to apply kernel_settings changes] ************************
task path: /home/rhack/system-role/devel/kernel_settings/handlers/main.yml:2

RUNNING HANDLER [linux-system-roles.kernel_settings : clear the kernel_settings_reboot_required flag] **********************************
task path: /home/rhack/system-role/devel/kernel_settings/handlers/main.yml:12

PLAY RECAP *****************************************************************************************************************************
10.0.137.196               : ok=24   changed=12   unreachable=0    failed=1    skipped=2    rescued=0    ignored=0


If you need to reproduce, just run test in this branch against rhel8.

richm:
> If you need to reproduce, just run test in this branch against rhel8.

which image did you use?  the rhel-8 and rhel-x CI tests are passing.  I've also tested locally against my rhel8 qcow2 cloud image and it is working.

Is the code that you pushed for this PR different than the latest code in your `master` branch?

Note that *Fedora* has a problem with bootloader cmdline settings.  I have added a note to the README to explain this - https://github.com/linux-system-roles/kernel_settings/pull/9 - and in the test that I added, I skip bootloader cmdline testing - https://github.com/linux-system-roles/kernel_settings/blob/master/tests/tests_simple_settings.yml#L9-L14

richm:
Ok.  I was able to reproduce.  The problem is that `quiet` is part of the built-in bootloader cmdline options, and the `kernel_settings` role cannot remove that.  That also needs to be added to the documentation.  If you want to see the default cmdline options, check `/proc/cmdline` before running the role, or use `grubby --info=ALL` to see the default bootloader config.

RHEmployee:
Hello. So I removed check for quick from test.
Please review.

richm:
@RHEmployee more changes are coming, thanks to your testing - https://github.com/linux-system-roles/kernel_settings/pull/15 - basically, we are not going to support setting bootloader cmdline settings.

tests/tests_change_settings.yml
```diff
@@ -0,0 +1,114 @@
...
+          - name: fs.file-max
+            value: 400000
+          - name: kernel.threads-max
+            state: absent
+        kernel_settings_sysfs:
+          - name: /sys/class/net/lo/mtu
+            value: 65000
+
+    - name: check sysfs after role runs
+      command: grep 65000 /sys/class/net/lo/mtu
```
richm, line 40: `grep -x 65000 /sys/class/net/lo/mtu` - `-x` so that you will match only `^65000$` and not `11165000000`

tests/tests_change_settings.yml
```diff
@@ -0,0 +1,114 @@
...
+
+    - name: check sysctl after role runs
+      shell: sysctl -n fs.file-max | grep 400000
```
richm, line 43: same here - use `grep -x 400000` for exact match

tests/tests_change_settings.yml
```diff
@@ -0,0 +1,114 @@
...
+
+    - name: check sysctl after role runs
+      shell: sysctl -n kernel.threads-max | grep -Lvq 29968
```
richm, line 46: use -x here too

tests/tests_change_settings.yml
```diff
@@ -0,0 +1,114 @@
...
+
+    - name: reboot the machine - see if settings persist after reboot
+      reboot:
+        test_command: tuned-adm active
+
+    - name: check sysctl after reboot
+      shell: sysctl -n fs.file-max | grep 400000
```
richm, line 53: -x

tests/tests_change_settings.yml
```diff
@@ -0,0 +1,114 @@
...
+
+    - name: check sysfs after reboot
+      command: grep 65000 /sys/class/net/lo/mtu
```
richm, line 56: -x

tests/tests_change_settings.yml
```diff
@@ -0,0 +1,114 @@
...
+
+    - name: show cmdline
+      command: cat /proc/cmdline
+
+    - name: check sysctl after reboot
+      shell: sysctl -n kernel.threads-max | grep -Lvq 29968
```
richm, line 62: -x

tests/tests_change_settings.yml
```diff
@@ -0,0 +1,114 @@
...
+            state: absent
+          - name: kernel.threads-max
+            state: absent
+
+    - name: reboot the machine - see if settings persist after reboot
+      reboot:
+        test_command: tuned-adm active
+
+    - name: check sysctl after reboot
+      shell: sysctl -n fs.file-max | grep -Lvq 400000
```
richm, line 82: -x

tests/tests_change_settings.yml
```diff
@@ -0,0 +1,114 @@
...
+      vars:
+        kernel_settings_sysctl:
+          - name: fs.file-max
+            value: 400001
+        kernel_settings_sysfs:
+          - name: /sys/class/net/lo/mtu
+            value: 60666
+
+    - name: check sysctl
+      shell: sysctl -n fs.file-max | grep -q 400001
```
richm, line 96: -x

tests/tests_change_settings.yml
```diff
@@ -0,0 +1,114 @@
...
+
+    - name: check sysfs after role runs
+      command: grep 60666 /sys/class/net/lo/mtu
```
richm, line 99: -x

tests/tests_change_settings.yml
```diff
@@ -0,0 +1,114 @@
...
+      include_role:
+        name: linux-system-roles.kernel_settings
+      vars:
+        kernel_settings_sysctl:
+          state: empty
+        kernel_settings_sysfs:
+          state: empty
+
+    - name: check sysctl
+      shell: sysctl -n fs.file-max | grep -Lvq 400001
```
richm, line 111: -x

tests/tests_change_settings.yml
```diff
@@ -0,0 +1,114 @@
...
+
+    - name: check sysfs after role runs
+      command: grep -Lqv 60666 /sys/class/net/lo/mtu
```
richm, line 114: -x

tests/tests_change_settings.yml
```diff
@@ -0,0 +1,114 @@
...
+
+    - name: Ensure kernel settings profile directory exists
+      file:
+        path: /etc/tuned/kernel_settings
+        state: directory
+        mode: 0755
+
+    - name: Generate a configuration for kernel settings
+      copy:
+        src: tuned/etc/tuned/basic_settings/tuned.conf
```
richm, line 16: You can't use this file anymore.  You'll need to use or create a file that doesn't have the bootloader cmdline settings in it.

RHEmployee: Hello. Thank you. Which file is it?

richm: > Hello. Thank you. Which file is it?

There is no such file that is the same as `tuned/etc/tuned/basic_settings/tuned.conf` but without the bootloader cmdline settings.  I suggest you create a new profile + file e.g. `tests/tuned/etc/tuned/simple_settings/tuned.conf` which is the same as `tests/tuned/etc/tuned/basic_settings/tuned.conf` but without the bootloader cmdline settings, and use that file in your test.

RHEmployee: Many, many thanks. Should be solved now.

tests/tuned/etc/tuned/basic_settings/tuned_change.conf
```diff
@@ -0,0 +1,18 @@
+# tuned configuration
```
richm, line 1: The intention is that `tests/tuned/etc/tuned/` mirrors exactly what would be under `/etc/tuned` in a real system, so that if you want a valid `/etc/tuned` for testing, you can just copy `tests/tuned/etc/tuned/`.  Please correct me if I'm wrong, but from what I understand about tuned, each profile has its own named directory under `/etc/tuned/`, and each profile directory has one single config file named `tuned.conf` for that profile.  I think a profile directory may contain other files like scripts, but only 1 config file.  So having `tests/tuned/etc/tuned/basic_settings/` with both `tuned.conf` and `tuned_change.conf` breaks that model.  
Therefore, I suggest you create a new profile directory `tests/tuned/etc/tuned/change_settings/` and move `tuned_change.conf` to `tests/tuned/etc/tuned/change_settings/tuned.conf`

I apologize that my comment https://github.com/linux-system-roles/kernel_settings/pull/7#discussion_r440868873 was not clear, and that I did not explain the purpose and contents of `tests/etc/tuned/`

PR #8, opened 2020-05-19
do not overwrite existing profile file

richm:
The role was overwriting the existing profile file.
Fixed by using the `template` module with `force: no` - if
the profile already exists, do not overwrite with the template


richm:
@marmolak - this is is the bug you found while working on https://github.com/linux-system-roles/kernel_settings/pull/7

PR #9, opened 2020-06-04
add NOTE about bootloader cmdline settings with Fedora Cloud images

richm:
There are a couple of issues with bootloader cmdline settings.

* Cannot remove/replace default settings built-in to bootloader

* There is currently a problem with managing bootloader cmdline settings with
Fedora Cloud images for Fedora 31, 32, and 33 rawhide.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,6 +7,23 @@ and derivatives (RHEL and CentOS) and Fedora.
 
 * `tuned` homepage - https://github.com/redhat-performance/tuned
 
+Supported Distributions
+-----------------------
+* RHEL-7+, CentOS-7+
+* Fedora
+
+Limitations
+-----------
+
+### Fedora Cloud images - managing bootloader cmdline settings
+
+There is currently a problem with managing bootloader cmdline settings with
+Fedora Cloud images for Fedora 31, 32, and 33 rawhide.  We suggest you first
+update all of the packages (e.g. `dnf -y update`) and reboot the system before
+attempting to use `kernel_settings_bootloader_cmdline`.  And even then, it
+still might not work - the current Fedora 33 rawhide cloud images do not work
+even with the suggestion above.
+
 ## Requirements
 
 This role requires an operating system which has the `tuned` package and
--- a/README.md
+++ b/README.md
@@ -15,6 +15,10 @@ Supported Distributions
 Limitations
 -----------
 
+### Default bootloader settings cannot be removed/replaced
+
+This role cannot remove or replace bootloader cmdline options added by default.  For example, on a machine that you have not run the role on, check `cat /proc/cmdline`.  This role cannot remove those parameters.  In some cases it may be able to override them - see [kernel command-line parameters](https://www.kernel.org/doc/html/latest/admin-guide/kernel-parameters.html) for more information.  This role is used primarily to manage bootloader cmdline parameters related to performance and security.  It isn't meant as a general purpose bootloader management role.  If you need to manage the default bootloader cmdline settings, use the tools provided by your distribution like the `systemd-boot`, `grub2-*` and `grubby` tools.
+
 ### Fedora Cloud images - managing bootloader cmdline settings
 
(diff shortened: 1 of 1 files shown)


pcahyna:
The firewall example README has [Supported Distributions](https://github.com/linux-system-roles/firewall#supported-distributions) and [Limitations](https://github.com/linux-system-roles/firewall#limitations) where this information would fit.

richm:
@RHEmployee

richm:
@RHEmployee - what bootloader cmdline parameters do you typically use for tuned testing?

richm:
@yarda @olysonek - is this already mentioned in the tuned documentation?

yarda:
> @yarda @olysonek - is this already mentioned in the tuned documentation?

@richm it's expected. With the BLS the kernel options can be specified on multiple places and in multiple variables. It's very hard to remove the options from the right place and the correct way. Tools like grubby simplifies the task by just expanding (flattening) all the options first and then modifying them in the flat list, so you could get something which is very different from your intention - i.e. flat list of options, instead of the options specified hierarchically and in variables. That's why Tuned supports only removal of options it added itself. In the Tuned doc there is written:
```
bootloader
  Adds options to the kernel command line. This plug-in supports only the GRUB 2 boot loader.
```
I think we should add explicit note about it, I added comment to the https://github.com/redhat-performance/tuned/pull/264 (which is an attempt to improve the bootloader documentation) regarding this.

richm:
How about this?  https://github.com/redhat-performance/tuned/pull/264/files#r436098647

yarda:
> How about this? https://github.com/redhat-performance/tuned/pull/264/files#r436098647

Thanks, I will update the PR. The PR is currently waiting on the doc team for more thorough review.

pcahyna:
@yarda 
> With the BLS the kernel options can be specified on multiple places and in multiple variables.

what are those places and variables? The only thing I see in https://systemd.io/BOOT_LOADER_SPECIFICATION/ is that "`options` ... may appear more than once in which case all specified parameters are used in the order they are listed." but this does not seem particularly difficult to handle.

yarda:
> @yarda
> 
> > With the BLS the kernel options can be specified on multiple places and in multiple variables.
> 
> what are those places and variables? The only thing I see in https://systemd.io/BOOT_LOADER_SPECIFICATION/ is that "`options` ... may appear more than once in which case all specified parameters are used in the order they are listed." but this does not seem particularly difficult to handle.

E.g. from what I currently know (there maybe more):
- for BLS options can appear more than once in the boot loader entry
- for BLS multiple grubenv variables can be referenced from the boot loader entry, the variables are defined in the grubenv, e.g.: "options $kernelopts1 $kernelopts2", $kernelopts1/2 can be grubenv variables
- probably one variable could reference other variable
- other options can be specified in /etc/default/grub, e.g. GRUB_CMDLINE_LINUX, GRUB_CMDLINE_LINUX_DEFAULT, ..., this could be also defined recursively
- other options can be in grub.cfg or grub-efi.cfg (e.g.the  set option=value).
- the options can be generated by scripts from the /etc/grub.d

Use case for Tuned was to add additional kernel options to the minimal set of options configured by Anaconda and not change the way how the current options are configured on the system.

richm:
closing this PR - kernel_settings does not support bootloader cmdline settings

RHEmployee:
> @RHEmployee - what bootloader cmdline parameters do you typically use for tuned testing?

I don't have nothing specific. Mostly, we do regression testing so depends on bug.
However, you proposed spectre flag which seems to work. So I can use it.

richm:
> > @RHEmployee - what bootloader cmdline parameters do you typically use for tuned testing?
> 
> I don't have nothing specific. Mostly, we do regression testing so depends on bug.
> However, you proposed spectre flag which seems to work. So I can use it.

I doesn't matter any more since we have decided to drop support for bootloader cmdline settings in the kernel_settings role.

PR #10, opened 2020-06-05
bootloader cmdline values must be strings

richm:
Convert bootloader cmdline values to strings.

### DIFF
--- a/library/kernel_settings.py
+++ b/library/kernel_settings.py
@@ -320,7 +320,7 @@ class BLCmdLine(object):
     @classmethod
     def escapeval(cls, val):
         """make sure val is quoted as in shell"""
-        return shlex_quote(val)
+        return shlex_quote(str(val))
 
     def __str__(self):
         vallist = []


richm:
@RHEmployee
PR https://github.com/linux-system-roles/kernel_settings/pull/11 depends on this one

pcahyna:
Is there a test for the problem this PR is solving? Is the problem that the value can be a number and before the change the code did not convert the number properly to a string?

pcahyna:
I suppose `value: 10001` in PR #11 shows the problem, but why do tests in PR #11 pass without this change?

richm:
> Is there a test for the problem this PR is solving?

Yes.

> Is the problem that the value can be a number and before the change the code did not convert the number properly to a string?

Yes.

richm:
> > I suppose `value: 10001` in PR #11 shows the problem,
> 
> Yes.
> 
> > but why do tests in PR #11 pass without this change?
> 
> Good question. The travis/tox tests fail. I guess somewhere in the code the exception is being caught and discarded. Investigating.

The answer is that there is no `tests/tests*.yml` test that uses `panic`.  Shall I add one?

richm:
> > > I suppose `value: 10001` in PR #11 shows the problem,
> > 
> > 
> > Yes.
> > > but why do tests in PR #11 pass without this change?
> > 
> > 
> > Good question. The travis/tox tests fail. I guess somewhere in the code the exception is being caught and discarded. Investigating.
> 
> The answer is that there is no `tests/tests*.yml` test that uses `panic`. Shall I add one?

If I add `panic` to the `tests/tests_simple_settings.yml` it does indeed fail:
```
TASK [linux-system-roles.kernel_settings : Apply kernel settings] **************
task path: /home/rmeggins/linux-system-roles/kernel_settings/tasks/main.yml:36
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: TypeError: expected string or bytes-like object
fatal: [/var/lib/libvirt/images/rhel8.qcow2]: FAILED! => {
    "changed": false,
    "rc": 1
}
```
with a big stacktrace.

How about this - what if I combine #10 and #11 into a single commit/PR **and** add `panic` to the bootloader cmdline values to test?

pcahyna:
Sorry, I did not realize that #11 uses a numeric value in a unit test, not in an integration one. Unifying the two PRs and adding an integration test is a great idea - I suppose the integration test can be done by changing or adding one value here: https://github.com/linux-system-roles/kernel_settings/blob/2b7a6d219ae23fe201e20138accd6ac3d1bc29c1/tests/vars/vars_simple_settings.yml#L21

Another question. You addressed integers, but what if a user passes a boolean to a parameter like `spectre_v2`? Should booleans be converted to "on"/"off" strings, or should booleans be simply unsupported? It also depends on whether the kernel command line is completely consistent in its use of "on" and "off" for boolean values. Given that `spectre_v2` is actually not a boolean setting - it accept more than "on" and "off", another possible value is "auto" - it could be better to simply not support booleans.

richm:
closing in favor of https://github.com/linux-system-roles/kernel_settings/pull/12

richm:
> Given that spectre_v2 is actually not a boolean setting - it accept more than "on" and "off", another possible value is "auto" - it could be better to simply not support booleans.

The problem is that if the user uses something like this:
```yaml
  - name: spectre_v2
    value: true
```
and it gets converted to `spectre_v2=True` e.g. user specifies an Ansible boolean value -> converted to a `bool` type when passed to the module - python `bool` can only be `True` or `False` - module uses python `str(boolvalue)` to convert this to the string `"True"` or `"False"`.  Not sure what to do about this other than document that values should be always passed as either YAML `int` or `string` values.

pcahyna:
My idea was to have explicit support for booleans and convert them to "on" and "off" strings instead of "True" and "False". Whether it is a good idea, especially for settings that are not inherently boolean, is debatable.
I was more concerned about
```yaml
  - name: spectre_v2
    value: on
```
i.e. when user forgets to quote `on` and the YAML parser helpfully converts it to a boolean value.

> document that values should be always passed as either YAML int or string values.

That's a valid approach. If so, this should be not only documented but actually checked in the code, because there is room for surprises.

richm:
> My idea was to have explicit support for booleans and convert them to "on" and "off" strings instead of "True" and "False". Whether it is a good idea, especially for settings that are not inherently boolean, is debatable.
> I was more concerned about
> 
> ```yaml
> kernel_settings_bootloader_cmdline:
>   - name: spectre_v2
>     value: on
> ```
> 
> i.e. when user forgets to quote `on` and the YAML parser helpfully converts it to a boolean value.
> 
> > document that values should be always passed as either YAML int or string values.
> 
> That's a valid approach. If so, this should be not only documented but actually checked in the code, because there is room for surprises.

https://github.com/linux-system-roles/kernel_settings/pull/13 - convert YAML bool automatically to `"on"`/`"off"` for bootloader cmdline values

PR #11, opened 2020-06-05
use panic=10001 instead of quiet in examples, tests

richm:
Some of the systems we use for testing have `quiet` as a built-in
bootloader cmdline value.  The kernel_settings role cannot remove
or replace built-in values.  Instead, use `panic` with a value
of `10001` as it is seldom used and applicable to all platforms.
This also gives us a non-string value to use for testing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -203,23 +203,25 @@ kernel_settings_bootloader_cmdline:
   - name: spectre_v2
     value: "off"
   - name: nopti
-  - name: quiet
+  - name: panic
+    value: 10001
   - name: splash
 ```
-will result in the values `spectre_v2=off nopti quiet splash` being added to,
-or replacing, the existing `bootloader` `cmdline` values.  If you want to
-remove whatever was there, and replace it with your specified values, use
-`previous: replaced` in the value list:
+will result in the values `spectre_v2=off nopti panic=10001 splash` being
+added to, or replacing, the existing `bootloader` `cmdline` values.  If you
+want to remove whatever was there, and replace it with your specified values,
+use `previous: replaced` in the value list:
 ```yaml
 kernel_settings_bootloader_cmdline:
   - previous: replaced
   - name: spectre_v2
     value: "off"
   - name: nopti
-  - name: quiet
+  - name: panic
+    value: 10001
   - name: splash
 ```
-then the values `spectre_v2=off nopti quiet splash` will *replace* the
+then the values `spectre_v2=off nopti panic=10001 splash` will *replace* the
 existing `cmdline` values, if any.  If you want to remove specific values, use
 `state: absent` on the settings you want to remove:
 ```yaml
@@ -229,9 +231,9 @@ kernel_settings_bootloader_cmdline:
   - name: nopti
     state: absent
 ```
-if the previous value was `spectre_v2=off nopti quiet splash`, the items
+if the previous value was `spectre_v2=off nopti panic=10001 splash`, the items
 `spectre_v2=off` and `nopti` will be removed from the list, and the final
-value will be `quiet splash`.
+value will be `panic=10001 splash`.
 
 ## Dependencies
 
@@ -262,7 +264,8 @@ The `tuned` package is required for the default provider.
       - name: spectre_v2
         value: "off"
       - name: nopti
-      - name: quiet
+      - name: panic
+        value: 10001
       - name: splash
   roles:
     - linux-system-roles.kernel_settings
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -58,7 +58,8 @@ kernel_settings_transparent_hugepages_defrag_state: null
 #   - name: spectre_v2
 #     value: "off"
 #   - name: nopti
-#   - name: quiet
+#   - name: panic
+#     value: 10001
 #   - name: splash
 kernel_settings_bootloader_cmdline: []
 
--- a/library/kernel_settings.py
+++ b/library/kernel_settings.py
@@ -186,7 +186,7 @@ kernel_settings:
 # delete the /sys/kernel/debug/x86/retp_enabled setting
 # completely remove the vm section
 # add the bootloader cmdline arguments spectre_v2=off nopti
-# remove the bootloader cmdline arguments quiet splash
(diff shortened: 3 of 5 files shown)


richm:
@RHEmployee

pcahyna:
I am curious: what is a "built-in bootloader cmdline value"? What does one do if one does not want `quiet` there?

richm:
> I am curious: what is a "built-in bootloader cmdline value"? What does one do if one does not want `quiet` there?

It depends on the platform and the tools provided by the platform to update bootloader settings: e.g. https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/system_administrators_guide/ch-working_with_the_grub_2_boot_loader#sec-Introduction_to_GRUB_2

Perhaps it was a mistake for the kernel_settings role to have any support at all for bootloader cmdline settings given the limitations of tuned in this respect.  I would not want users to expect that the kernel_settings role is a fully-fledged bootloader management role.  The main reason why kernel_settings has support for some sort of bootloader cmdline management is that these settings are often used in conjunction with or as a replacement for setting values using `sysctl` or setting values in `/sys/fs`.  But it would be very difficult to explain to a user who didn't already know something about this what parameters/settings those are.  Perhaps the tuned documentation explains this - @yarda @olysonek?

So if we can't figure out a way to explain this in the kernel_settings role in a satisfactory manner, I think the alternative is to remove the bootloader cmdline functionality from the kernel_settings role, and instead provide a fully-fledged bootloader management role - @tabowling ?

pcahyna:
> > I am curious: what is a "built-in bootloader cmdline value"? What does one do if one does not want `quiet` there?
> 
> It depends on the platform and the tools provided by the platform to update bootloader settings: e.g. https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/system_administrators_guide/ch-working_with_the_grub_2_boot_loader#sec-Introduction_to_GRUB_2

I must say that I don't see anything there that says that removing `quiet` from the kernel command line should be impossible or that it is a "built-in bootloader cmdline value". Nevertheless,

> Perhaps it was a mistake for the kernel_settings role to have any support at all for bootloader cmdline settings given the limitations of tuned in this respect. I would not want users to expect that the kernel_settings role is a fully-fledged bootloader management role. The main reason why kernel_settings has support for some sort of bootloader cmdline management is that these settings are often used in conjunction with or as a replacement for setting values using `sysctl` or setting values in `/sys/fs`. 

There is no straightforward mapping between `/sys` or sysctl parameters and the kernel command line parameters, or is it?

> But it would be very difficult to explain to a user who didn't already know something about this what parameters/settings those are. Perhaps the tuned documentation explains this - @yarda @olysonek?
> 
> So if we can't figure out a way to explain this in the kernel_settings role in a satisfactory manner, I think the alternative is to remove the bootloader cmdline functionality from the kernel_settings role, and instead provide a fully-fledged bootloader management role - @tabowling ?

In general I agree, if the functuionality is not ready, we can omit it from the first release. Whether to add it later to this role or create a new role is a question that can be solved later and does not seem to be the main problem (if there is some problem implementing it, the same problem will occur regardless of whether it is done in a separate role or this role).

That said, "It depends on the platform and the tools provided by the platform" convinces me that there is a value of providing a platform-independent abstraction, because depending on platform-specific tools is exactly what the users want to avoid. So in the longer term we need this functionality somewhere.

richm:
> I must say that I don't see anything there that says that removing quiet from the kernel command line should be impossible or that it is a "built-in bootloader cmdline value".

Of course there won't be anything in there for some value like "quiet".  You asked "What does one do if one does not want quiet there?" and instead of giving the actual sequence of `grub-*` or `grubby` commands, I pointed you to the manual for managing bootloader settings for one of our supported platforms.  I suppose I should have explained what I was doing.

> There is no straightforward mapping between /sys or sysctl parameters and the kernel command line parameters, or is it?

No, there is not.  This is one of those things that the user will have to know how to do based on their research into kernel bootloader options.  https://github.com/linux-system-roles/kernel_settings/pull/9 includes a link to the kernel cmdline parameters - https://www.kernel.org/doc/html/latest/admin-guide/kernel-parameters.html

> In general I agree, if the functuionality is not ready, we can omit it from the first release.

ok but there are some kernel_settings cases such as spectre/meltdown performance mitigation where having the ability to set sysctl, sysfs, *and* bootloader cmdline settings in one place is convenient.  Also, I'm hoping that perhaps tuned has some elegant way of explaining this in the documentation that I can borrow.

pcahyna:
> > I must say that I don't see anything there that says that removing quiet from the kernel command line should be impossible or that it is a "built-in bootloader cmdline value".
> 
> Of course there won't be anything in there for some value like "quiet". You asked "What does one do if one does not want quiet there?" and instead of giving the actual sequence of `grub-*` or `grubby` commands, I pointed you to the manual for managing bootloader settings for one of our supported platforms. I suppose I should have explained what I was doing.

Yes, and surprisingly, the manual actually *does* say something about `quiet`. it uses it as an example of a value that can be removed: https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/system_administrators_guide/sec-Making_Persistent_Changes_to_a_GRUB_2_Menu_Using_the_grubby_Tool#bh-Adding_and_Removing_Arguments_from_a_GRUB_2_Menu_Entry 
"To add and remove arguments from a kernel's GRUB 2 menu entry, use a command as follows:
```
~]# grubby --remove-args="rhgb quiet" --args=console=ttyS0,115200 --update-kernel /boot/vmlinuz-3.10.0-229.4.2.el7.x86_64
```
"
And it does not mention any platform limitations.

So, when you say that the kernel_settings role cannot remove or replace `quiet`, I was surprised to see referenced here a document which uses this very argument as an example of something that can be removed.

Anyway, thanks to @yarda 's comment https://github.com/linux-system-roles/kernel_settings/pull/9#issuecomment-639689407 I now understand better the reason for this limitation and why you probably should not use the approach described in that document inside the role. Seems to be a difficult problem indeed.

richm:
closing in favor of https://github.com/linux-system-roles/kernel_settings/pull/12

PR #12, opened 2020-06-05
use panic instead of quiet in blcmd examples, tests

richm:
Some of the systems we use for testing have `quiet` as a built-in
bootloader cmdline value.  The kernel_settings role cannot remove
or replace built-in values.  Instead, use `panic` with a value
of `10001` as it is seldom used and applicable to all platforms.
This also gives us a non-string value to use for testing.
There was a bug with using non-string values in blcmdline
values.  This commit also fixes that bug.

### DIFF
--- a/README.md
+++ b/README.md
@@ -203,23 +203,25 @@ kernel_settings_bootloader_cmdline:
   - name: spectre_v2
     value: "off"
   - name: nopti
-  - name: quiet
+  - name: panic
+    value: 10001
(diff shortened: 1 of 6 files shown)


richm:
This PR includes both #10 and #11 and adds `panic` to `tests/tests_simple_settings.yml` - you can decide if you would rather have this single commit/PR or separate commits/PRs.

PR #13, opened 2020-06-05
replace bootloader cmdline bool values with on/off

richm:
The use of `on` and `off` for bootloader cmdline values is
quite common - https://www.kernel.org/doc/html/latest/admin-guide/kernel-parameters.html

Even though we will strongly encourage users to use a YAML string
value for bootloader cmdline settings, it will help users avoid
mistakes if we convert YAML boolean values to `"on"`/`"off"`.

For example, if the user types in this instead of `value: "on"`:
```yaml
  - name: someparam
    value: on
```
we will convert this to `someparam=on`.  This also means values
like:
```yaml
  - name: someparam
    value: true
```
will also be converted to `someparam=on`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -235,6 +235,21 @@ if the previous value was `spectre_v2=off nopti panic=10001 splash`, the items
 `spectre_v2=off` and `nopti` will be removed from the list, and the final
 value will be `panic=10001 splash`.
 
+**NOTE**: Use either string values or numeric values to avoid ambiguity or errors with boolean values.  For example, a value like `value: on` will be converted to the string value `"True"` in the stored configuration, since `on` is a YAML boolean value for `True`.  Instead, use `value: "on"` to mean the actual string `"on"`. For bootloader cmdline values, since `on` and `off` are commonly used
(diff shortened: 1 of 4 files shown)


pcahyna:
Another approach would be to disallow booleans altogether and fail if a boolean is passed as a command line option value, forcing the user to use a proper string - less intuitive perhaps, but safer. @yarda @olysonek what do you think?

pcahyna:
The argument against the proposed conversion and for disallowing booleans completely is that one can not be sure whether converting a boolean to "on"/"off" or to "yes"/"no" or to 0/1 is right. Looking at the kernel parameters document, some settings actually use "yes"/"no", like `cmo_free_hint`. Given that we don't know what convention is used for a particular setting, I am afraid that we should not do this conversion, because it would transform a user-supplied `yes` into the string "on" and that's not what the user is expecting, either :-( 
Some of the settings actually are documented as: "Format: <bool>  (1/Y/y=enable, 0/N/n=disable)".

richm:
closing PR in favor of https://github.com/linux-system-roles/kernel_settings/pull/14

PR #14, opened 2020-06-05
Do not allow bool typed values

richm:
Do not allow [YAML bool type](https://yaml.org/type/bool.html) values.

For example, if the user types in this instead of `value: "on"`:
```yaml
  - name: someparam
    value: on
```
the role will issue an error.


richm:
> lgtm. Is there a well-defined standard for representing booleans in sysctl? If it is better defined than for command line settings, we could allow booleans there. But this can be done in the future, it is better to accept a more restricted set of values and expand it later than vice-versa.

Good question.  Unfortunately, it seems more complicated than I assumed :-(

The documentation for the `sysctl` module doesn't say anything about handling boolean values: https://docs.ansible.com/ansible/latest/modules/sysctl_module.html#sysctl-module

However, if you look at the code, there is some very special purpose handling of boolean values: https://github.com/ansible/ansible/blob/stable-2.9/lib/ansible/modules/system/sysctl.py#L212

That is
- if `value is True`, it will be converted to the string value `'1'`
- if `value is False`, it will be converted to the string value `'0'`
- if `value` is one of the strings in this set https://github.com/ansible/ansible/blob/stable-2.9/lib/ansible/module_utils/parsing/convert_bool.py#L8 it will be converted to the string value `'1'`
- if `value` is one of the strings in this set https://github.com/ansible/ansible/blob/stable-2.9/lib/ansible/module_utils/parsing/convert_bool.py#L9 it will be converted to the string value `'0'`

However, the tuned sysctl plugin https://github.com/redhat-performance/tuned/blob/master/tuned/plugins/plugin_sysctl.py does not do anything special with the values e.g. try to convert `'yes'` to `'1'` - all of the values are string values as read from the profile file, so there is no conversion of `bool` type to `'1'` or `'0'`.

If we want kernel_settings to work like the Ansible `sysctl` module, we need to convert the values in a similar fashion - if the user specifies `value: yes`, this is passed to the module as the boolean value `True` which should be converted to a string `'1'`.  If the user specifies `value: "yes"`, the module converts this to a string `'1'`.  Based on my limited knowledge of `sysctl`, this makes sense to me, as almost every `sysctl` parameter is a number, and where for boolean parameters, `0` means to disable it, and `1` means to enable it.  The very few string valued parameters seem to have quite meaningful values such as `NONE` or a MAC address or one of a limited set of choices.  e.g.
```
net.ipv4.tcp_allowed_congestion_control = reno cubic
net.ipv4.tcp_available_ulp = espintcp mptcp
kernel.version = #1 SMP Thu May 14 23:26:14 UTC 2020
kernel.seccomp.actions_avail = kill_process kill_thread trap errno user_notif trace log allow
```

tl;dr - it seems like the Ansible `sysctl` is "doing the right thing" and the kernel_settings `kernel_settings_sysctl` parameter should behave the same way.

Also note that this only applies to the `kernel_settings_sysctl` if we want it to behave like the Ansible `sysctl` module - there is no such convention for `kernel_settings_sysfs` or the other parameters.

richm:
@pcahyna this is what the role looks like when using the `sysctl` to implement `kernel_settings_sysctl` - wdyt?

pcahyna:
> If we want kernel_settings to work like the Ansible sysctl module

Compatibility with the Ansible sysctl module is a good thing, but I would say not 100% required. We can leave it (boolean conversion) as a feature to be implenteted later.

> this is what the role looks like when using the Ansible sysctl module to implement kernel_settings_sysctl - wdyt?

I would prefer not to do such a rapid change in course in the last moment. Wasn't the argument for using tuned to set sysctl parameters that tuned is a single place to set them, while when using sysctl.d files, tuned profiles and sysctl.d files might fight over the control of a setting?

PR #18, opened 2020-06-27
do not use omit - does not work on nested values with ansible 2.7

richm:
The use of `omit` with second level or lower values does not
work on ansible 2.7.  The keys are passed to the module with
the value `__omit_token_xxxx`.  Instead, use the value `none`
or some other value which is empty appropriate for the key being
set, and have the module strip all empty values.

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -41,41 +41,34 @@
       - name: "{{ 'cpu_affinity'
                   if kernel_settings_systemd_cpu_affinity
                   or kernel_settings_systemd_cpu_affinity_state
-                  else omit }}"
-        value: "{{ kernel_settings_systemd_cpu_affinity
-                   if kernel_settings_systemd_cpu_affinity
-                   else omit }}"
-        state: "{{ kernel_settings_systemd_cpu_affinity_state
-                   if kernel_settings_systemd_cpu_affinity_state
-                   else omit }}"
+                  else none }}"
+        value: "{{ kernel_settings_systemd_cpu_affinity |
+                   d(none, true) }}"
+        state: "{{ kernel_settings_systemd_cpu_affinity_state |
+                   d(none, true) }}"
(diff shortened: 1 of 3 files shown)


richm:
@nhosoi

richm:
[citest]

tasks/main.yml
```diff
@@ -41,41 +41,34 @@
...
-                  else omit }}"
-        value: "{{ kernel_settings_systemd_cpu_affinity
-                   if kernel_settings_systemd_cpu_affinity
-                   else omit }}"
-        state: "{{ kernel_settings_systemd_cpu_affinity_state
-                   if kernel_settings_systemd_cpu_affinity_state
-                   else omit }}"
+                  else none }}"
+        value: "{{ kernel_settings_systemd_cpu_affinity |
+                   d(none, true) }}"
```
nhosoi, line 46: May I ask why the default value is `none, true` for all?

richm: `none` in Ansible/jinja2 is converted to the python `None` value - the module treats this value as empty, and removes it.  I could have used any "empty" value, such as `""` or `[]` or `{}` - but `none` seems clearer.

richm: also - `kernel_settings_systemd_cpu_affinity | d(none, true)` - https://jinja.palletsprojects.com/en/2.11.x/templates/#default

the `true` means that the `default` filter will evaluate the given variable in the boolean context if defined - it is a shorthand for `if kernel_settings_systemd_cpu_affinity` - so I can do the `is defined` and "evaluates to true in a boolean context" test in one filter

nhosoi:
lgtm

richm:
@pcahyna I have confirmed that this is a problem only for ansible 2.7 - it works fine with ansible 2.8 - it interprets `omit` correctly for the nested settings

PR #19, opened 2020-06-29
RHELPLAN-46757 - bug with current implementation of platform/version …

nhosoi:
…specific vars/tasks file include/import

Updated based on the oasis-roles meta_standards:
  Supporting multiple distributions and versions
  - Always specify the explicit, absolute path to the files to be included
  - Using loop instead of with_first_found.

### DIFF
--- a/tests/tasks/assert_kernel_settings.yml
+++ b/tests/tasks/assert_kernel_settings.yml
@@ -1,10 +1,14 @@
 ---
-- name: Set variables used by assert/testing code
-  include_vars: "{{ item }}"
-  with_first_found:
-    - "vars/tests_{{ ansible_distribution }}_\
-      {{ ansible_distribution_major_version }}.yml"
-    - "vars/tests_default.yml"
+- name: Set version specific variables
+  include_vars: "{{ lookup('first_found', ffparams) }}"
+  vars:
+    ffparams:
+      files:
+        - "tests_{{ ansible_distribution }}_\
+          {{ ansible_distribution_major_version }}.yml"
+        - "tests_default.yml"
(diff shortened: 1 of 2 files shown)


tests/tasks/assert_kernel_settings.yml
```diff
@@ -1,10 +1,18 @@
...
+      files:
+        - "{{ ansible_facts['distribution'] }}_\
+          {{ ansible_facts['distribution_version'] }}.yml"
+        - "{{ ansible_facts['distribution'] }}_\
+          {{ ansible_facts['distribution_major_version'] }}.yml"
+        - "{{ ansible_facts['distribution'] }}.yml"
+        - "{{ ansible_facts['os_family'] }}.yml"
+        - "default.yml"
+      paths:
+        - "{{ role_path }}/vars"
```
richm, line 15: `role_path` is only defined inside the role - this test code runs outside of the role
I don't know if there is a way to get the current working directory as an absolute path - for test code, I think it is ok to use a relative path, especially since the file name already has "tests_" in it - so something like
```yaml

- name: Set version specific variables
  include_vars: "{{ lookup('first_found', ffparams) }}"
  vars:
    ffparams:
      files:
        - "tests_{{ ansible_distribution }}_\
          {{ ansible_distribution_major_version }}.yml"
        - "tests_default.yml"
    paths:
      - vars
```

nhosoi:
*sigh* I just should have copied and paste...

nhosoi:
[citest bad]

nhosoi:
Thanks, @richm. Merging.

PR #28, opened 2020-08-25
use import_role instead of include_vars for role vars

richm:
The test verification needs access to the role private vars from
vars/main.yml.  Using include_vars requires a path which may be
different depending on the packaging.  Instead, rely on a "trick"
with import_role - using `when: false` will include the role, skipping
the explicit tasks, but running the implicit tasks such as loading
the vars and defaults.

### DIFF
--- a/tests/tests_simple_settings.yml
+++ b/tests/tests_simple_settings.yml
@@ -17,9 +17,13 @@
       include_role:
         name: linux-system-roles.kernel_settings
 
-    - name: include vars from role to use for verification
-      include_vars:
-        file: roles/linux-system-roles.kernel_settings/vars/main.yml
+    # NOTE: This is only to include the vars from vars/main.yml - the
+    # when: false will skip all of the tasks in the role except the
+    # implicit tasks to load the vars and defaults
+    - name: import the role to include vars to use for verification
+      import_role:
+        name: linux-system-roles.kernel_settings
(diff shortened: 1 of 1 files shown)


nhosoi:
Great! If we could avoid using include_vars in the kernel_settings test scripts, I can get rid of this code. ;)

https://github.com/linux-system-roles/auto-maintenance/pull/13/files#diff-86af87396754accdaa23d5ba76b59052R356-R362

tests/tests_simple_settings.yml
```diff
@@ -17,9 +17,13 @@
...
-    - name: include vars from role to use for verification
-      include_vars:
-        file: roles/linux-system-roles.kernel_settings/vars/main.yml
+    # NOTE: This is only to include the vars from vars/main.yml - the
+    # when: false will skip all of the tasks in the role except the
+    # implicit tasks to load the vars and defaults
+    - name: import the role to include vars to use for verification
+      import_role:
+        name: linux-system-roles.kernel_settings
+      when: false
```
pcahyna, line 26: Since you are including the role in the task immediately above, can't you get rid both of the `include_role` above and of `when: false` and do everything you need in `import_role`? Or add `public` to `include_role` above: https://docs.ansible.com/ansible/latest/modules/include_role_module.html#parameter-public

PR #29, opened 2020-08-25
use include_role with public: true to get role vars

richm:
The test playbook needs the role private vars to verify the settings.
use include_role with `public: true` so that those vars will be
exported and available to the tests
@nhosoi @pcahyna

### DIFF
--- a/tests/tests_simple_settings.yml
+++ b/tests/tests_simple_settings.yml
@@ -13,17 +13,14 @@
         __kernel_settings_check_reboot: false
       when: ansible_distribution == "Fedora"
 
+    # use public: true here so that the private role
+    # variables will be exported - we use
+    # __kernel_settings_profile_filename to verify
+    # that the settings were applied correctly
     - name: apply the settings - call the role
       include_role:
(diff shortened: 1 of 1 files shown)


tests/tests_simple_settings.yml
```diff
@@ -16,14 +16,7 @@
     - name: apply the settings - call the role
       include_role:
         name: linux-system-roles.kernel_settings
-
-    # NOTE: This is only to include the vars from vars/main.yml - the
-    # when: false will skip all of the tasks in the role except the
-    # implicit tasks to load the vars and defaults
```
pcahyna, line 22: I would prefer to keep a comment explaining why `public: true` is needed (what variable from the role do we need later).

richm: fixed in https://github.com/linux-system-roles/kernel_settings/pull/29/commits/4f14c744e9988aca05d72e4cf0c9ae3b7e29a459

pcahyna (approved):
lgtm now

PR #49, opened 2021-03-02
Fix ansible-test sanity issues

richm:
There were many issues related to formatting of the module docs, import
order, lint issues, and more. This also fixes an issue with
install_tuned_for_testing.sh, and gets rid of the unused run_pylint.sh


nhosoi (approved):
lgtm

PR #51, opened 2021-03-12
RHELPLAN-69928  Add a note to each module Doc to indicate it is private

nhosoi:
@richm, this is an example of sentence to start the synopsis of module doc. If you could suggest what to put, I'm going to add it to each module. Thanks!

### DIFF
--- a/library/kernel_settings.py
+++ b/library/kernel_settings.py
@@ -25,6 +25,7 @@ short_description: Manage kernel settings using tuned via a wrapper
 version_added: "2.8"
 
 description:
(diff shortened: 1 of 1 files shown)


library/kernel_settings.py
```diff
@@ -25,6 +25,7 @@
 version_added: "2.8"
 
 description:
+    - This is a private module which is not supported for the public use.
```
richm, line 28: `WARNING: Do not use this module directly!  It is only for role internal use.`

richm (approved):
lgtm

PR #61, opened 2021-07-28
disable bootloader tests - fails on rhel9

richm:
Disable testing bootloader config - this is failing on rhel9 - the
role doesn't support bootloader settings anyway - this will make
testing cleaner.

See also https://bugzilla.redhat.com/show_bug.cgi?id=1944599

### DIFF
--- a/library/kernel_settings.py
+++ b/library/kernel_settings.py
@@ -784,7 +784,7 @@ def remove_if_empty(params):
 
 
 def run_module():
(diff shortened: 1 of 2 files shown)


richm:
NOTE: The rhel-x failures are due to tuned-2.16 not being available yet

nhosoi (approved):
lgtm

PR #77, opened 2022-02-24
ensure python-configobj is present

richm:
tuned recently switched from configobj to configparser.  The role cannot
yet switch to use configparser.  Ensure that the package python-configobj
is present.

### DIFF
--- a/vars/CentOS_7.yml
+++ b/vars/CentOS_7.yml
@@ -1,0 +1,3 @@
+---
+__kernel_settings_packages: ["tuned", "python-configobj"]
+__kernel_settings_services: ["tuned"]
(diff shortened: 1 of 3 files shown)


nhosoi (approved):
lgtm

vars/CentOS_7.yml
```diff
@@ -0,0 +1,3 @@
+---
+__kernel_settings_packages: ["tuned", "python-configobj"]
+__kernel_settings_services: ["tuned"]
```
spetrosi, line 3: Why do you create new vars files for CentOS_7 and RedHat_7?

richm: The package is named `python3-configobj` on every platform except EL7.

spetrosi (approved):
lgtm

vars/CentOS_7.yml
spetrosi: Oh yes, I am blind

PR #78, opened 2022-02-28
add sysctl for testing - package procps-ng

richm:
Looks like rhel 8.6 dropped `procps-ng` recently, so no `sysctl` command for test playbooks - add it back

### DIFF
--- a/tests/tests_change_settings.yml
+++ b/tests/tests_change_settings.yml
@@ -1,8 +1,9 @@
 - hosts: all
   tasks:
+    # procps-ng for sysctl cmd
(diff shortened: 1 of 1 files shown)


nhosoi (approved):
lgtm

I could be wrong, but `network` also uses sysctl in its testing? And it does not add `procps-ng`?

richm:
> lgtm
> 
> I could be wrong, but `network` also uses sysctl in its testing? And it does not add `procps-ng`?

Then it will likely fail on rhel-8-y . . .

PR #79, opened 2022-04-13
support gather_facts: false; support setup-snapshot.yml

richm:
Some users use `gather_facts: false` in their playbooks.  This changes
the role to work in that case, by gathering only the facts it requires
to run.
CI testing can be sped up by creating a snapshot image pre-installed
with packages.  tests/setup-snapshot.yml can be used by a CI system
to do this.

### DIFF
--- a/tasks/set_vars.yml
+++ b/tasks/set_vars.yml
@@ -1,0 +1,22 @@
+---
+- name: Ensure ansible_facts used by role
+  setup:
+    gather_subset: min
+  when: not ansible_facts.keys() | list |
+    intersect(__kernel_settings_required_facts) ==
+    __kernel_settings_required_facts
+
+- name: Set platform/version specific variables
+  include_vars: "{{ lookup('first_found', ffparams) }}"
+  vars:
+    ffparams:
+      files:
+        - "{{ ansible_facts['distribution'] }}_\
+          {{ ansible_facts['distribution_version'] }}.yml"
+        - "{{ ansible_facts['distribution'] }}_\
+          {{ ansible_facts['distribution_major_version'] }}.yml"
+        - "{{ ansible_facts['distribution'] }}.yml"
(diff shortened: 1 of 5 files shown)


tasks/set_vars.yml
```diff
@@ -0,0 +1,22 @@
...
+      files:
+        - "{{ ansible_facts['distribution'] }}_\
+          {{ ansible_facts['distribution_version'] }}.yml"
+        - "{{ ansible_facts['distribution'] }}_\
+          {{ ansible_facts['distribution_major_version'] }}.yml"
+        - "{{ ansible_facts['distribution'] }}.yml"
+        - "{{ ansible_facts['os_family'] }}.yml"
+        - "default.yml"
+      paths:
+        - "{{ role_path }}/vars"
```
nhosoi, line 22: Out of curiosity, why you don't use `loop` and `__vars_file` here as you usually do in the other `set_vars.yml` like this?
https://github.com/linux-system-roles/certificate/blob/master/tasks/set_vars.yml#L8-L21

richm: With the certificate role, it was safe to convert the first_found method to the "is file" method - every para.  For other roles, it didn't seem safe, so I didn't convert.  I probably should not have converted certificate like this - it is not related to the PR - I thought it would simplify porting the change to other roles - but other roles are not so simple.

richm: In the case of the kernel_settings role, it is safe to convert the first_found method to the "is file" method - but I think we should not make this change in order to support `gather_facts: false` and setup-snapshot.yml

nhosoi: Thanks for the details, @richm. I am currently working on `timesync` which shares the same issue. I'm not going to convert it to `loop`, then. Thanks!

Issue #90, opened 2022-07-18
Setting kernel_settings_reboot_ok to true never reboots the server


timgriffiths:
I believe there is a bug in here, 

https://github.com/linux-system-roles/kernel_settings/blob/master/tasks/main.yml#L115

This line I believe should not be in the when clause, 

`- not kernel_settings_reboot_ok | d(false)`

or should be 

`- kernel_settings_reboot_ok | d(false)`

otherwise, the handler will never fire

richm:
Thanks - please try https://github.com/linux-system-roles/kernel_settings/pull/93
note that the role only sets reboot_required if setting bootloader settings which isn't really supported https://github.com/linux-system-roles/kernel_settings/blob/master/library/kernel_settings.py#L449
We are hoping to have an actual bootloader (grub, uefi, etc.) at some point in the future - https://github.com/linux-system-roles/bootloader

richm:
Fixed by #93

PR #93, opened 2022-07-25
Set the kernel_settings_reboot_required when reboot needed

richm:
Previously the role would only set `kernel_settings_reboot_required`
if the user did not specify `kernel_settings_reboot_ok: true`.
The role will now set `kernel_settings_reboot_required` whenever
the system needs a reboot, and rely on the handler to clear the flag
if the user has set `kernel_settings_reboot_ok: true`.

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -112,4 +112,3 @@
     kernel_settings_reboot_required: true
   when:
     - __kernel_settings_register_module.reboot_required | d(false)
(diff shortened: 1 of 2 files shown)


richm:
fixes https://github.com/linux-system-roles/kernel_settings/issues/90

richm:
[citest]

nhosoi (approved):
lgtm

Issue #95, opened 2022-07-31
lookup plugin: first_found not found when using ansible-core 2.12.2  


mto79:
```
TASK [fedora.linux_system_roles.kernel_settings : Set platform/version specific variables] ************************************************
fatal: [xxxxxxxx]: FAILED! => {"msg": "An unhandled exception occurred while running the lookup plugin 'first_found'. Error was a <class 'KeyError'>, original message: 'Requested entry (plugin_type: lookup plugin: first_found setting: files ) was not defined in configuration.'. 'Requested entry (plugin_type: lookup plugin: first_found setting: files ) was not defined in configuration.'"}
```

richm:
> ```
> TASK [fedora.linux_system_roles.kernel_settings : Set platform/version specific variables] ************************************************
> fatal: [xxxxxxxx]: FAILED! => {"msg": "An unhandled exception occurred while running the lookup plugin 'first_found'. Error was a <class 'KeyError'>, original message: 'Requested entry (plugin_type: lookup plugin: first_found setting: files ) was not defined in configuration.'. 'Requested entry (plugin_type: lookup plugin: first_found setting: files ) was not defined in configuration.'"}
> ```

What platform is your control node (i.e. where you installed ansible)?  What platform is the managed node?  How did you install ansible-core?  pip?  rpm? deb?

mto79:
Hi Richard,

The platform for the control node is: Red Hat Enterprise Linux release 8.6 (Ootpa)
The platform for the managed node is also: Red Hat Enterprise Linux release 8.6 (Ootpa)

Ansible-core is comming from the offical repo (so rpm based):

ansible-core-2.12.2-3.1.el8.x86_64 : SSH-based configuration management, deployment, and task execution system
Repo        : rhel-8-for-x86_64-appstream-rpms
Matched from:
Provide    : ansible-core = 2.12.2-3.1.el8

ansible [core 2.12.2]
  config file = /etc/ansible/ansible.cfg
  configured module search path = ['/root/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python3.8/site-packages/ansible
  ansible collection location = /root/.ansible/collections:/usr/share/ansible/collections
  executable location = /usr/bin/ansible
  python version = 3.8.12 (default, Sep 16 2021, 10:46:05) [GCC 8.5.0 20210514 (Red Hat 8.5.0-3)]
  jinja version = 2.11.3
[...]

PR #96, opened 2022-07-31
Pull request for issue 95

mto79:
In Ansible 2.9.13 there are already namespaces so suggestion for the refactoring of this system role, so to use already anisble.builtin.first_found for now. 

Tested it and worked. 
 
It is so minor that users with higher ansible versions can use this system-role anyway for the time being. 
Before  officialy updating the code for ansibe-core 2.12.2

### DIFF
--- a/tasks/set_vars.yml
+++ b/tasks/set_vars.yml
@@ -7,7 +7,7 @@
     __kernel_settings_required_facts
 
 - name: Set platform/version specific variables
-  include_vars: "{{ lookup('first_found', ffparams) }}"
+  include_vars: "{{ lookup('ansible.builtin.first_found', ffparams) }}"
   vars:
     ffparams:
       files:


richm:
This is really odd.  In addition to the questions I have in https://github.com/linux-system-roles/kernel_settings/issues/95:

* The guidance I have received from the Ansible core team is that FQCN for ansible builtins should not be required for the foreseeable future.
* 2.12.2 is not the latest version of ansible-core 2.12.x - can you reproduce the original problem with the latest 2.12.x?

There are several other built-in filters/tests that are used - they do not have this problem?
```
# tox -e ansible-plugin-scan
...
The following ansible.builtin modules are used at runtime:
command debug fail file include_tasks include_vars package reboot service set_fact setup shell slurp template
The following ansible.builtin filters are used at runtime:
b64decode comment intersect regex_replace
The following ansible.builtin tests are used at runtime:
changed failed
The following ansible.builtin lookups are used at runtime:
lookup
...
```
(I'll note that first_found is not in this list - which is perhaps a clue as to why that particular filter has this problem)

mto79:
I will try to reproduce with latest 2.12.x of ansible. (This requires repo sync in satelitte, so will come back latter for the result)

PR #143, opened 2023-04-07
ansible-lint - use changed_when even if using conditional

richm:
ansible-lint now requires the use of `changed_when` even if the
command already uses `when`


richm:
[citest]

PR #220, opened 2024-08-09
fix: Use tuned files instead of using it as a module

richm:
The previous version of the kernel_settings role used `tuned` as
a python library, and had a `kernel_settings` module which was a
wrapper around this code.  However, `tuned` version 2.23 has changed
its internal API and it is no longer possible to use it as a python
library.  Instead, the kernel_settings role has been refactored to
read/write `tuned` config files, and let the `tuned` daemon manage
the settings.

In addition, `tuned` 2.23 changed the location of the profile directory,
so the kernel_settings role will now determine the location of the
profile directory depending on the `tuned` version.

The old `kernel_settings` module is removed, along with all of the
python unit testing code.

A new `kernel_settings_get_config` module has been created which will
simply parse and return the given config file as a `dict`.


tasks/main.yml
```diff
@@ -33,6 +43,32 @@
...
+  set_fact:
+    __kernel_settings_profile_parent: "{{ __prof_from_conf
+      if __prof_from_conf | length > 0 else
+      __kernel_settings_tuned_dir ~ __prof_subdir }}"
+  vars:
+    __data: "{{ __kernel_settings_register_tuned_main.data }}"
+    __prof_from_conf: "{{ __data.get('profile_dirs', '').split(',')[-1] }}"
+    __prof_subdir: "{{ '/profiles'
+      if __kernel_settings_tuned_version is version('2.23.0', '>=')
+      else '' }}"
```
spetrosi, line 70: You can set __kernel_settings_tuned_version here too, it's not used anywhere else

richm:
[citest]

tasks/main.yml
```diff
@@ -46,96 +82,139 @@
...
-    force: false
-    mode: "0644"
+- name: Get active_profile
+  slurp:
+    path: "{{ __kernel_settings_tuned_active_profile }}"
+  register: __kernel_settings_tuned_current_profile
+
+- name: Set active_profile
+  set_fact:
+    __kernel_settings_active_profile: "{{ __active_profile }}"
```
spetrosi, line 92: You can use what is set for `__active_profile` here directly

tasks/main.yml
```diff
@@ -46,96 +82,139 @@
...
+  copy:
+    content: >
+      {{ __kernel_settings_active_profile }}
+    dest: "{{ __kernel_settings_tuned_active_profile }}"
+    mode: preserve
+  register: __kernel_settings_register_profile
+
+- name: Set profile_mode to manual
+  copy:
+    content: >
```
spetrosi, line 110:
```suggestion
    content: manual
```
is it the same thing?

tasks/main.yml
```diff
@@ -46,96 +82,139 @@
...
+      manual
```
spetrosi, line 111:
```suggestion
```

tasks/main.yml
```diff
@@ -46,96 +82,139 @@
...
+    __trans_defrag_old: "{{
+      __kernel_settings_profile_contents.data.get('vm', {}).get('transparent_hugepage.defrag')
+      if not kernel_settings_purge
+      and kernel_settings_transparent_hugepages_defrag != __kernel_settings_state_absent
+      else none }}"
+    __sections:
+      - name: sysctl
+        new: "{{ kernel_settings_sysctl | difference([__kernel_settings_previous_replaced]) | list
+          if kernel_settings_sysctl != __kernel_settings_state_empty
+          else [] }}"
```
spetrosi, line 156: Why __sysctl_old is set to dict `{}` and new is set to list `[]`?

tasks/main.yml
```diff
@@ -46,96 +82,139 @@
...
 
 - name: Tuned apply settings
-  command: >
-    tuned-adm profile '{{ __kernel_settings_register_module.active_profile }}'
-  when: __kernel_settings_register_module is changed  # noqa no-handler
+  command: >-
+    tuned-adm profile {{ __kernel_settings_active_profile | quote }}
+  when:
+    - not __kernel_settings_register_profile is changed
+    - not __kernel_settings_register_mode is changed
```
spetrosi, line 196: In the above task, there is `or` between these two checks, here they both should be `changed` to restart apply tuned?

tasks/main.yml
richm: not sure what you mean

tasks/main.yml
richm: I want to ensure the file ends with a newline - not sure if `content: manual` will ensure that

tasks/main.yml
spetrosi: I mean do not set `__active_profile` in vars, just paste what `__active_profile` equals right in `__kernel_settings_active_profile`.

tasks/main.yml
richm: `new` is a `list` of `dict` - basically the same format as passed in by `kernel_settings_sysctl` et. al. - see https://github.com/linux-system-roles/kernel_settings/pull/220/files#diff-d5995b3130f3bd6a8890b9832f7250dbaf05507afd6e49f7e371152a459c50abR8

`old` is a `dict` as returned by `kernel_settings_get_config` and processed here - https://github.com/linux-system-roles/kernel_settings/pull/220/files#diff-d5995b3130f3bd6a8890b9832f7250dbaf05507afd6e49f7e371152a459c50abR15

richm:
[citest]

PR #295, opened 2026-02-25
fix: ensure idempotence by writing settings in sorted order

richm:
Cause: On EL7, the underlying module uses an unordered dict, so the order of the settings being
written can change between role invocations.

Consequence: The role may report changed when nothing actually changed except the order.

Fix: The settings are sorted before being written so that the order is consistent.

Result: The role does not report changed when nothing has changed.

## Summary by Sourcery

Bug Fixes:
- Prevent spurious changes being reported by the role by sorting kernel settings before rendering them in the template.


richm:
[citest]

spetrosi (approved):

PR #300, opened 2026-03-19
test: ensure role gathers the facts it uses by having test clear_facts before include_role

richm:
The role gathers the facts it uses.  For example, if the user uses
`ANSIBLE_GATHERING=explicit`, the role uses the `setup` module with the
facts and subsets it requires.

This change allows us to test this.  Before every role invocation, the test
will use `meta: clear_facts` so that the role starts with no facts.

Create a task file tests/tasks/run_role_with_clear_facts.yml to do the tasks
to clear the facts and run the role.  Note that this means we don't need to
use `gather_facts` for the tests.

Some vars defined using `ansible_facts` have been changed to be defined with
`set_fact` instead.  This is because of the fact that `vars` are lazily
evaluated - the var might be referenced when the facts have been cleared, and
will issue an error like `ansible_facts["distribution"] is undefined`.  This is
typically done for blocks that have a `when` condition that uses `ansible_facts`
and the block has a role invocation using run_role_with_clear_facts.yml
These have been rewritten to define the `when` condition using `set_fact`.  This
is because the `when` condition is evaluated every time a task is invoked in the
block, and if the facts are cleared, this will raise an undefined variable error.

## Summary by Sourcery

Ensure kernel_settings role tests run the role with cleared Ansible facts to verify it gathers required facts explicitly.

New Features:
- Add a reusable test task file to clear Ansible facts and invoke the kernel_settings role with configurable options.

Enhancements:
- Refactor existing kernel_settings tests to use a shared task file for invoking the role instead of including the role directly, removing explicit fact gathering from playbooks.

### DIFF
--- a/tests/tasks/cleanup.yml
+++ b/tests/tasks/cleanup.yml
@@ -7,8 +7,7 @@
 - name: Use role purge to remove settings
   block:
     - name: Run role with purge to remove everything
-      include_role:
-        name: linux-system-roles.kernel_settings
+      include_tasks: tasks/run_role_with_clear_facts.yml
       vars:
         kernel_settings_purge: true
         kernel_settings_sysctl: []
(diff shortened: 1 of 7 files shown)


richm:
[citest]

